PR #16, opened 2016-05-23
Annotate all new non-empty English CoreNLP stories

### LINKED ISSUE #15: Enable CoreNLP annotation for all new English language stories
_(trying to post again, GitHub is having difficulties today)_

As we seem to have a capability to annotate all new English language stories with CoreNLP (#12), we should start doing that.

The task will probably include moving existing annotations to S3 too. An average annotation takes up ~4 KB of disk space, so annotating all the existing English language stories would require 0,5+ TB of database disk space.

I think the best way to annotate old stories would be to finally deploy RabbitMQ as a job broker and then stuff all the old stories into the message queue with a lower priority.

pypt:
Fixes #15.

Should I merge & deploy it, or does @rahulbot has to do some changes on his side?


hroberts:
I don't think any of this will change anything on rahul's side.  He is still querying for specific media sets, so he won't even start getting more stories.  When he wants to turn geotagging on for the whole stream, he can change his query to just search for english language sentences.

pypt:
So I'll then just enable CoreNLP processing for all new English language stories.

PR #22, opened 2016-06-14
Remove topics, media sets, dashboards, etc.

pypt:
@hroberts, could you have a look at this cleanup, verify that I accidentally didn't delete anything that's currently being used?

Summary:

* deleted old dashboard code
* deleted old topics code, making way for the upcoming controversies -> topics rename
* deleted old queries code
* deleted mostly everything under `root/public_ui/` as there's no more public UI apparently
* deleted `MediaWords::I18N` translations as they weren't being used by anything except the public UI
* deleted some helpers in 12a0d09, not sure if you were using those
* **dropped `old_download_time`, `old_state`, `file_status`, `relative_file_path` columns from `downloads` table** (1b079ba). The first two weren't doing anything useful, `file_status` was probably related to that disk crash some time ago but with downloads now being stored on S3 wasn't carrying any useful information, and `relative_file_path` had something to do with Tar archives
* removed "solr_facets_wrapper_server", whatever that was
* removed some other obsolete scripts, no-op unit tests, moved some code around

If everything's cool, I'll go make a push request to Rahul so that his unit tests continue to be working after deploying this.


hroberts:
I just scanned through it for obvious wrong chunks, but I didn't find anything.  Still very possible there are issues with dependencies with existing code.

PR #169, opened 2017-06-28
Chinese language support

natlungfy:
… enabled zh support in Language.pm


pypt:
Hi Natalie, good PR! Could you please take care of the following technicalities, and then right afterwards we could merge + deploy?

* Replace list of punctuation characters in `punc` with some other means of punctuation / symbol removal:
	
	It's pretty unsafe to have a finite list of symbols to be removed from the text as you don't really know what kind of punctuation you will encounter in the text.
	
	The way the symbol character removal is done in Japanese language module is that the Japanese language analyzer provides part of speech marks for every token (including punctuation), and we allow through only tokens of a certain part of speech. Does Jieba provide such a functionality?
	
	If not, maybe it would be worth it to create a helper `is_punctuation()` in `mediawords.util.text` and filter out punctuation [using Unicode character classes](https://stackoverflow.com/questions/11066400/remove-punctuation-from-unicode-formatted-strings/11066687#11066687) (or some other way, I don't know). If you choose to go this path, make sure to write an unit test for this function, e.g. test it against ASCII punctuation (`.`), Unicode punctuation (`。` - ideographic full stop, `～` - full width tilde), etc.

* `lib/MediaWords/Languages/t/zh.t` is still failing for whatever reason, I think you forgot to change some expected / actual strings somewhere.

	Perl unit test for the language module (`zh.t`) should not be anywhere near extensive, it's just a small sanity test to ensure that Perl-Python integration works as expected.

* Use static `__jieba` instance in the class instead of `jieba`.
	
	Right now you're importing `jieba` module and initializing its own tokenizer singleton with `jieba.initialize()`. It is generally a good practice to keep the class dependencies self-contained when it is possible and avoid having / using singletons and other "global" stuff, because, for example, if some other part of the code decided to use `jieba` too and re-initialize it again, unexpected stuff might happen.
	
	So, you probably want to initialize `self.__jieba = jieba.Tokenizer()` in `__init__` (maybe do `from jieba import Tokenizer as JiebaTokenizer` to make it clear which tokenizer is it), and then use `self.__jieba` instead of `jieba` everywhere.

* Duplicate code in defining paths to `dict.txt.big` and `userdict.txt`: `os.path.join(self.__dict_path, 'dict.txt.big')`, `os.path.join(self.__dict_path, 'userdict.txt')`

* Remove unused imports at the top of the file (`import jieba.posseg as pseg`, `from typing import Dict`, `from mediawords.util.text import random_string`, `jieba.analyse`).

* Remove `__jieba` in `TestChineseTokenizer` as it doesn't seem to be used.

* Rename `_dict_path` to `__dict_path` in `McChineseTokenizer`. The comment states that the variable is used in the unit test but that is not the case, so it should be made private (even though Python doesn't have protected / private variables so [it's only a convention](https://stackoverflow.com/a/1641236/200603)).

* Where did you take `dict.txt.big` from? Could you add a README or some other reference near this dictionary?

* Same goes for Chinese stopword list `zh_stopwords.txt`. Could you please provide a source in the comment section (stopword file parser ignores strings starting with `# `)?

* In `Language.pm`, remove:

		# Chinese disabled because of poor word segmentation
		
	comment as it's no longer the case.

* Remove `# -*- coding: UTF-8 -*-` in `zh.py` and `test_zh.py`. Python 3 source files are UTF-8 by default, so this comment is not needed.

* Remove `zh_dict.txt` from the old Chinese implementation.

* Reformat Python files according to PEP 8; it's just a [single shortcut on PyCharm](https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#format-code-in-pep-8).

* Make your code [have no warnings](https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#write-code-with-no-warnings), e.g. unused imports, bad spacing, etc. PyCharm would also a great tool here.


**Some side notes:**

* I've noticed that you've committed and then reverted some debugging `print()` statements.
	
	While "old style" debugging using `print()` is generally unavoidable, the great thing about IDEs like PyCharm is that they have debuggers which, when used together with test-driven development, make it very easy to find out what is happening in your code by simply setting a couple of breakpoints in your unit test and then running it.
	
	If you haven't had a look into [PyCharm's debugger](https://www.jetbrains.com/help/pycharm/debugging.html), I'd suggest you invest some time and effort into figuring it out as it will save you tremendous amounts of head scratching in the future no matter what you do.

* Admittedly, we're pretty far from "practice what you preach" when it comes to commit messages, but out of general curiosity, skim through the legendary [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) page.

	I find that putting in more effort into my own commit messages becomes more and more important for me over time, as I constantly have to revisit my own code after months or years, so when committing some changes, I try to think of the "future me" (or some other developer) when trying to explain what my code is all about.
	
	One last strategy that I try to employ to do my future-self a favor is that I try to make my [commits atomic](https://www.freshconsulting.com/atomic-commits/), i.e. I avoid having "fix stuff" commits with changes over 10kLoC. While it takes more time to do all those separate commits, it:
	
	* allows me to `git revert` stuff,
	* makes testing easier - one can then pinpoint the test failure to a specific change in the code; also, you can make a small commit, push it to [Travis's continuous integration](https://travis-ci.org/berkmancenter/mediacloud) and continue working while waiting for Travis to tell you whether or not something breaks,
	* makes thinking about own code more systematic, i.e. the need to have atomic commits forces me to think about the implementation in the means of discrete steps that I should take to do something, and I find it's harder to make various mistakes and forget things that way.

Looking forward to your small fixes, and can't wait to merge + deploy and have a functional Chinese language support!

pypt:
> Duplicate code in defining paths to `dict.txt.big` and `userdict.txt`: `os.path.join(self.__dict_path, 'dict.txt.big')`, `os.path.join(self.__dict_path, 'userdict.txt')`

Some clarification: currently you're defining the full path to both dictionaries twice:

```python
if not os.path.isfile(os.path.join(self._dict_path, 'dict.txt.big')):
    raise ...

if not os.path.isfile(os.path.join(self._dict_path, 'userdict.txt')):
    raise ...

try:
    # <...>
    jieba.set_dictionary(os.path.join(self._dict_path,'dict.txt.big'))
    jieba.load_userdict(os.path.join(self._dict_path,'userdict.txt'))
except Exception as ex:
    raise ...
```

and I would suggest for those paths to be assigned to variables first:

```python
jieba_dict_path = os.path.join(self._dict_path, 'dict.txt.big')
jieba_userdict_path = os.path.join(self._dict_path, 'userdict.txt')

if not os.path.isfile(jieba_dict_path):
    raise ...
if not os.path.isfile(jieba_userdict_path):
    raise ...

try:
    # <...>
    jieba.set_dictionary(jieba_dict_path)
    jieba.load_userdict(jieba_userdict_path)
except Exception as ex:
    raise ...
```

so if you choose to change paths to one or more of the dictionaries, there would be only a single line to change.

Maybe those two variables (`jieba_dict_path` and `jieba_userdict_path`) would be better off as static class variables (just like `_dict_path` is) as then all the hardcoded paths would be to be found at a single place (at the top of the class definition).

pypt:
Hey @natlungfy, how is it going? If you have some intermediate changes already, consider pushing them as separate commits!

natlungfy:
@pypt Sorry for the delay! I have made the changes but got stuck on the is_punctuation function as it is causing some word tokenization tests to fail. I tested it separately by feeding it a sentence with all kinds of Chinese punctuation and it worked, but fails when I run it within the Media Cloud project. I'll push the rest of my changes now.

pypt:
Merged and deployed.

PR #170, opened 2017-06-29
Create token_pool.py

Alan32Liu:
tokenize articles

### DIFF
--- a/mediacloud/mediawords/util/topic_modeling/token_pool.py
+++ b/mediacloud/mediawords/util/topic_modeling/token_pool.py
@@ -1,0 +1,104 @@
+from mediawords.db import connect_to_db
+import json
+import re
+# from stop_words import get_stop_words
+
+
+class TokenPool:
+    """ Fetch the sentences and break it down to words.
+    """
+    DB_QUERY = """SELECT stories_id, sentence FROM story_sentences"""
+    STOP_WORDS = "lib/MediaWords/Languages/resources/en_stopwords.txt"
+    DELIMITERS = "[^\w]"
+
+    def __init__(self):
+        """Initialisations"""
+        pass
+
+    def fetch_sentences(self):
+        """
+        Fetch the sentence from DB
+        :return: the sentences in json format
+        """
+        db_connection = connect_to_db()
+        sentences_hash = db_connection.query(self.DB_QUERY).hashes()
+        sentences_json = json.loads(s=json.dumps(obj=sentences_hash))
+        db_connection.disconnect()
+
+        return sentences_json
+
+    def tokenize_sentence(self, sentences):
+        """
+        Break the sentence down into tokens and group them by article ID
+        :param sentences: a json containing sentences and their article id
+        :return: a dictionary of articles and words in them
+        """
+        articles = {}
+
+        for sentence in sentences:
+            if sentence['stories_id'] not in articles.keys():
+                articles[sentence['stories_id']] = []
+            articles[sentence['stories_id']] \
+                += self.eliminate_symbols(article_sentence=sentence['sentence'])
+
+        return articles
+
+    def eliminate_symbols(self, article_sentence):
+        """
+        Remove symbols in the given list of words in article
+        :param article_sentence: a sentence in an article
+        :return: a list of non-symbol tokens
+        """
+        return re.split(pattern=self.DELIMITERS, string=article_sentence)
+
+    def fetch_stopwords(self):
+        """
+        Fetch the stopwords from file en_stopwords.txt
+        :return: all stopwords in the file
+        """
+        stopwords = [element[:-1] for element in open(self.STOP_WORDS).readlines()]
+        return stopwords
+
+    def eliminate_stopwords(self, article_words):
+        """
+        Remove stopwords in the given list of words in article
+        :param article_words: a list containing all words in an article
+        :return: a list of all the meaningful words
+        """
+        stopwords_file = self.fetch_stopwords()
+        # stopwords_package = get_stop_words('en')
+
+        stemmed_tokens_via_file = [word for word in article_words
+                                   if ((len(word) > 1) and (word.lower() not in stopwords_file))]
+
+        # stemmed_tokens_via_package = [word for word in article_words
+        #                               if ((len(word) > 1)
+        #                                   and (word.lower() not in stopwords_package))]
+
+        # print(set(stemmed_tokens_via_file) - set(stemmed_tokens_via_package))
+        # print(set(stemmed_tokens_via_package) - set(stemmed_tokens_via_file))
+
+        return stemmed_tokens_via_file
+
+    def output_tokens(self):
+        """
+        Go though each step to output the tokens of articles
+        :return: a dictionary with key as the id of each article and value as the useful tokens
+        """
+        sentences = self.fetch_sentences()
+        tokens = self.tokenize_sentence(sentences=sentences)
+        stemmed_tokens = {}
+
+        counter = 0
+        for article_id, article_token in tokens.items():
+            stemmed_tokens[article_id] = self.eliminate_stopwords(article_words=article_token)
+            counter += 1
+            if counter > 4:
+                break
+
+        return stemmed_tokens
+
+
+# A sample output
+# pool = TokenPool()
+# print(pool.output_tokens().popitem())


Alan32Liu:
1. Two LDA model (with different package, not sure which one is better yet)
2. A path helper to assist import
3. modified token_pool to make it compatible with LDA models

pypt:
Hey Dongge, thanks for the PR! Good job on encapsulating both LDA libraries into classes! Also, it's great that you've put in effort into docstring'ing your methods - as you might have noticed, we don't write docstrings for our methods (but I suppose we should).

Before you continue working on LDA, I'd like to merge this PR into `master` branch. When working in a team, it is a common practice to integrate and deploy code as often as possible to prevent separate branches from diverging too far from whatever's running in production. Merging and deploying now would allow us to maybe make use of your code already, plus it might give you some extra confidence in your work too!

Before the merge, I'd like you to consider making some additional changes in your branch. Some of these changes are common "good" practices used in software development, some others are just ways we like to do things. I hope that you won't feel overwhelmed by the huge list of changes.


**General:**

* Write at least some basic sanity unit tests.

    As detailed in [coding guidelines](https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#write-unit-tests), it is a good practice to test your code automatically; even more so in your case, as you'll be likely away after the summer, and we'll be left with your code which, when untested, is pretty hard to maintain.

    I suppose the unit tests should cover at least:

    * token pool -- sentence tokenization (with `How do you do?` parameter, does one get `['How', 'do', 'you', 'do']`; also, what happens with sentences non-breaking prefixes like *I enjoy cartoons, e.g. the ones drawn by Dr. Seuss.*?), stopword removal (do stopwords get removed as they should?)

    * LDA -- with a given list of stories (or single story tokens), do we get the expected list of suggested topics?

* Unsure why do you need `path_helper.py` at all. I've deleted it and everything works fine for me.

* Refactor sample calls of classes (e.g. `model = ModelLDA(); print(model.summarize())`) into unit tests, or use the `if __name__ == "__main__":` trick it you want to leave them.

    Python [executes code when it `import`s the module](https://stackoverflow.com/a/419185/200603), so if we chose to `import` your modules from somewhere else, we would run sample calls to modules without us knowing about it happening.

    As I understand, you use those calls to test your code. I think the best thing to do here would be for you to write unit tests for every thing that you would like to test, and then run those unit tests repeatedly while working on your code. This practice is called test-driven development (TDD for short), it basically means that one should treat the unit tests not as a formality to be added to the code after completing it, but as a useful tool to be used while doing the initial development.

    If you would like to leave your sample calls in the code, use the `if __name__ == "__main__":` trick detailed in the link above.

* Reformat code with [PyCharm's keyboard shortcut](https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#format-code-in-pep-8).

* Declare function parameter and return types.

    Python 3's type hinting is great! Here's the [excerpt from the coding guidelines](https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#declare-function-parameter-and-return-value-types) to help you with it.

* Decouple `token_pool` and `model_*` modules, I think they shouldn't know about each other.

    As LDA models simply spit out a list of topics from a list of terms, I don't think they should be doing any database access themselves. Separating those concerns (fetching + tokenizing stories, and guessing topics from tokenized stories) would make it easier to cover both functions with unit tests too.


**Token pool:**

* Use NLTK in word tokenization.

    Tokenizing text into tokens is not that easy as removing non-`\w` symbols via regexp as is the case in your current implementation. Consider the following sentence:

        I enjoy cartoons, e.g. the ones drawn by Dr. Seuss.

    If you chose to remove symbols (or split the sentence into tokens by `\s`), `e.g.` and `Dr. Seuss` would get split into separate tokens (`e`, `g`; `Dr`, `Seuss`), which is not what you want.

    What's commonly done in this case is maintaining a list of [non-breaking prefixes for every language](https://github.com/berkmancenter/mediacloud/blob/master/lib/MediaWords/Languages/resources/en_nonbreaking_prefixes.txt). However, I'd suggest you don't dig too deep into sentence tokenization and just reuse whatever [NLTK has to provide at this point](http://www.nltk.org/api/nltk.tokenize.html).

* Why are you serializing sentences to `json` in `token_pool.py` instead of just returning a `list`?

* I think you forgot to stem the terms. I'd suggest you use [NLTK's English language Porter stemmer](http://www.nltk.org/howto/stem.html) for that.

* Make methods not used elsewhere private (prepend `__`, e.g. rename `fetch_sentences()` to `__fetch_sentences()`) or at least protected (prepend a single `_`).

* You probably want to get a list of stories and their sentences, not a quasi-random list of sentences that belong to random stories.  Also, it is likely that you want only English language stories and have them in the order in which they appeared in the story. So, please reuse the SQL query (or a variant of it) that I've sent you in an email titled "Sample dump, database tables, tokenization and stopwords":

    ```sql
    SELECT stories_id,
           sentence
    FROM stories
        INNER JOIN story_sentences
            ON stories.stories_id = story_sentences.stories_id
    WHERE stories.language = 'en' -- some stories are not in English
    ORDER BY stories.stories_id,
             story_sentences.sentence_number
    ```

* Prepend `mc_root_path()` to `STOP_WORDS`:

    ```python
    STOP_WORDS = os.path.join(mc_root_path(), "lib/MediaWords/Languages/resources/en_stopwords.txt")
    ```

* Make `db` a parameter instead of reconnecting with `connect_to_db()` every time.

    The unit test should `connect_to_db()` for you and pass the database handler to the `TokenPool` helper.


**LDA:**

* Add `gensim`, `lda` to [`requirements.txt`](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/requirements.txt). Both are third-party libraries and need to be installed somehow. We install dependencies using Pip, which uses `requirements.txt`.

* `import gensim` only once; use `gensim.corpora` and `gensim.models` afterwards.

* Make `ModelLDA` and `ModelLDA2` into concrete subclasses of an abstract class that defines interface, with an abstract method `summarize()` expecting a list of parameters and returning predictable return values.

* Rename `ModelLDA` and `ModelLDA2` to something more meaningful, maybe `ModelGenSim` and `ModelLDA`?

* Move initialization stuff to class constructors.

    On every call to `summarize()`, you're re-initializing a couple of resources which (I think) would be better off if made into instance variables and initialized in a constructor.

* Reduce procedural code by adding more parameters to class methods instead of having them just "do the magic thing" and return something:
    * Make `summarize()` accept a list of stories and their tokens as a parameter instead of having them fetch those stories themselves via `TokenPool`.
    * Eliminate `STORY_NUMBER`, method should probably do its job with all the stories passed to it.
    * Make `TOPIC_NUMBER` into a parameter (maybe with a default value) in order to make it easier to vary the number of topics to be generated using the code
    * Make `WORD_NUMBER` into a parameter (probably with a default value too).
    * Make number of iterations configurable everywhere (via parameter to constructor / `summarize()`).
    * What is `RANDOM_STATE`?

* Passing arguments in an array with "magic" indexes (`stemmed_tokens[0]`, `stemmed_tokens[1]`) is to be avoided because it's not apparent what is being held as values under those indexes (e.g. is `stemmed_tokens[0]` a story ID or a list of tokens?). Please reorganize those into an object or find some other solution.

* How do we interpret the return value of `summarize()`?

    `ModelLDA` returns:

    ```python
    {16: [(0, '0.043*"migrants" + 0.017*"Rio" + 0.017*"Grande" + 0.015*"asylum"')], 11: [(0, '0.053*"Utah" + 0.032*"couples" + 0.021*"ban" + 0.016*"Judge"')], 13: [(0, '0.130*"Sebelius" + 0.087*"Kathleen" + 0.087*"marred" + 0.087*"resigning"')], 14: [(0, '0.027*"Republicans" + 0.024*"Obama" + 0.024*"Senate" + 0.022*"Republican"')], 15: [(0, '0.024*"league" + 0.021*"lawsuit" + 0.014*"hockey" + 0.014*"violence"')]}
    ```

    `ModelLDA2` returns:

    ```python
    {16: ['Grande', 'asylum', 'immigration', 'smugglers'], 17: ['rice', 'Persian', 'fried', 'sabzi'], 18: ['Lerner', 'Republican', 'conservative', 'contempt'], 19: ['journalists', 'judge', 'Greste', 'Jazeera'], 20: ['TV', 'media', 'Twitter', 'Facebook'], 21: ['Colbert', 'CBS', 'Letterman', 'Moonves'], 11: ['Utah', 'couples', 'ban', 'plaintiffs'], 13: ['Lerner', 'Republican', 'conservative', 'contempt'], 14: ['Obama', 'Republicans', 'Senate', 'Bush'], 15: ['league', 'lawsuit', 'violence', 'complaint']}
    ```

    Given that both methods are supposed to achieve the same thing, they should have identical return values.

* Why does `summarize()` operate only on a single story per iteration in `model_lda.py` and on all stories in `model_lda2.py`?

* How will it work with millions of stories (the terms of which won't fit into RAM)? Can you make both models into objects that could be trained iteratively, i.e.:

    ```python
    x = ModelLDA()
    x.add(story1_tokens)
    x.add(story2_tokens)
    x.add(story3_tokens)
    # <...>
    print(x.topics())
    ```


**Side notes:**

The following is admittedly some copy-pasta from another pull request that I have reviewed recently :), but I hope that you will find it useful too:

Admittedly, we're pretty far from "practice what you preach" when it comes to commit messages, but out of general curiosity, skim through the legendary [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) page.

I find that putting in more effort into my own commit messages becomes more and more important for me over time, as I constantly have to revisit my own code after months or years, so when committing some changes, I try to think of the "future me" (or some other developer) when trying to explain what my code is all about.

One last strategy that I try to employ to do my future-self a favor is that I try to make my [commits atomic](https://www.freshconsulting.com/atomic-commits/), i.e. I avoid having "fix stuff" commits with changes over 10kLoC. While it takes more time to do all those separate commits, it:

* allows me to `git revert` stuff,
* makes testing easier - one can then pinpoint the test failure to a specific change in the code; also, you can make a small commit, push it to [Travis's continuous integration](https://travis-ci.org/berkmancenter/mediacloud) and continue working while waiting for Travis to tell you whether or not something breaks,
* makes thinking about own code more systematic, i.e. the need to have atomic commits forces me to think about the implementation in the means of discrete steps that I should take to do something, and I find it's harder to make various mistakes and forget things that way.

---

Very promising work so far, Dongge, looking forward to your fixes!

pypt:
1. Please sort the newly added dependencies in `requirements.txt` as asked in the header of the file.

2. You mention in `requirements.txt` that one needs to install NLTK data but you don't do install it anywhere in the install scripts, so this is one of the reasons why the Travis build is failing.

    The general idea of doing those Travis builds is to ensure that everything what needs to be installed gets installed by the setup scripts, and that all the unit tests pass with whatever gets installed by Travis. Each Travis build is run in a "clean room" (a virtual machine that gets created just for the occassion), so you have to install everything yourself with a setup script to make your unit tests work.

    So, you probably want to add something like:

    ```bash
    echo "Installing NLTK data..."
    if [ `uname` == 'Darwin' ]; then
        NLTK_DATA_PATH=/usr/local/share/nltk_data
    else
        NLTK_DATA_PATH=/usr/share/nltk_data
    fi
    $COMMAND_PREFIX python$PYTHON3_MAJOR_VERSION -m nltk.downloader all -d "$NLTK_DATA_PATH"
    ```

    at the end of `install/install_python_dependencies.sh`.

    Some notes about what's happening in the lines above:

    * [Installing NLTK data](http://www.nltk.org/data.html#command-line-installation) page says that the recommended system location for NLTK data is different on OS X and Linux
    * We can't use `nltk.download()` because we have no GUI on Travis nor production, so we just install everything.
    * `COMMAND_PREFIX` is set to either `sudo` (on Linux) or an empty string (on OS X) because `/usr/local/share/nltk_data` is user-writable on OS X and `/usr/share/nltk_data` needs a privileged user access on Linux.)
    
    You probably want to add `/usr/share/nltk_data` (Travis runs on Ubuntu) to `cache/directories/` in file `.travis.yml` too; it takes a while to install all the NLTK data so we might as well cache it after installing once. Not sure if this is going to work though, but worth a try.

3. Separate LDA model unit tests and token pool tests from the database.

    Another reason why Travis builds are failing is that the database on Travis doesn't contain any data. On your local machine, you've imported the test dump that I've sent you and thus the tests on your system pass because there's data present in PostgreSQL that you're running. PostgreSQL on Travis doesn't contain any data by default (unless you load some data into PostgreSQL before you do testing), so that's why your tests are failing.

    On Perl, we have a `MediaWords::Test::DB::test_on_test_database` helper that, when used, preloads the database with a schema; then we write some data to PostgreSQL, run a subroutine that is being tested, compare the results, and throw this test database away. Unfortunately, this helper hasn't been ported to Python yet, but there's no need for you to use it. Instead, please test the individual helper methods (sentence tokenization, stopword removal) with some text data hardcoded into the test, e.g.:


    ```python
    def test_tokenize_sentence_into_words():
        assert tokenize_sentence_into_words('Dongge is in Australia.') == ['Dongge', 'is', 'in', 'Australia']
        # ...
    
    def test_remove_stopwords():
        assert remove_stopwords(['Dongge', 'is', 'in', 'Australia']) == ['Dongge', 'Australia']
        # ...
    ```

    This would:

    1. Break the dependency of your tests from the database.
    2. Make tests faster.
    3. Allow you to test out various input values (`None`, empty strings, true positives, false negatives, ...) more easily.

    Please apply the same approach to your model tests too.

4. Remove `path_helper.py`.

    No idea why your PyCharm is unable to resolve modules.

    1. Are you using PyCharm?
    2. Do you use Media Cloud's virtualenv in PyCharm (*Preferences* -> *Project: mediacloud* -> *Project Interpreter*; interpreter should be `3.5.x virtualenv at ~/path/to/mc-venv`)?
    3. Is the file `mc-venv/lib/python3.5/site-packages/mediacloud.pth` present on your system, and do the contents of that file point to `../../../../mediacloud/`?

5. Move tests to the same directory as the modules that are being tested, as per our [coding guidelines](https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#keep-unit-tests-in-the-same-directory-prefix-file-names-with-test_).

6. Get rid of warnings (there are some in `test_model_gensim.py` and `test_model_lda.py`)

7. *(from last PR review)* Why are you serializing sentences to `json` in `token_pool.py` instead of just returning a `list`?

    Right now you seem to serialize the database results to JSON and then unserialize it right afterwards. How come?

9. If `_eliminate_symbols()` no longer does symbol elimination but tokenizes sentences to words, I think it should have an appropriate name too.

10. Keep table names in the SQL queries instead of having them in constants.

    One has to read and modify the SQL query more often than the table / column names change, so it's not worth storing them in separate variables (which are used only once anyway) and sacrifice simplicity and readability that way.

11. *(from last PR review)* Please use the SQL query from the last PR review.

    I'm not sure if your SQL query does what you want it to do. It reads from `story_sentences`, then joins `stories`, then joins `story_sentences` again? Also, ordering by `sentence_number` is missing too. If my provided SQL query doesn't work for you, please let me know.

12. You wrote me some nice and useful explanations in how both of your model approaches work. I think it would be great to have them in docstrings and / or comments.

Alan32Liu:
# Status of Last Week
- [x] Apply nonnegative matrix factorization (NMF) model and compare its performance with existing models

- [x] Write test cases for NMF

- [x] Sort the newly added dependencies in `requirements.txt`

- [x] Add WordNet 
    - Add script to `install/install_python_dependencies.sh`
    - [x] add `/usr/share/nltk_data` to `.travis.yml`
    - [ ] It failed on travis, because it cannot find module `nltk` in line: `$COMMAND_PREFIX python$PYTHON3_MAJOR_VERSION -m nltk.downloader all -d "$NLTK_DATA_PATH"`, even though `nltk` was successfully installed before this step. Now sure why this is happening yet, but I will try to figure it out.

- [x] Separate `test_model_*.py` from database.
    - [x] To mimic the output of database query, I stored a small part of the database in `sample_stories.txt` in the same format as database outputs (i.e. `List[Dict[int, str]]`). The constructor of `TokenPool` can now take one of two things, either a database handler (to be used formally) or a TextIOWrapper (to be used in tests)
    - [x] I did the same to `test_token_pool.py`

- [x] Remove `path_helper.py`
    - It is not caused by PyCharm but Python3.5. I am coding with PyCharm and compiling with the command line.
    - Yes, interpreter is correct
    - Yes, that that file points to the correct directory

- [x] Move tests to the same directory as the modules. 

- [x] Get rid of warnings. It has three kinds of warnings
    - [x] Suppressed the warning caused by function `lemmatize()` of `nltk`. I understand that I should not suppress warnings, but this warning is caused by an unclosed file in this built-in method, so I am not sure if there is anything else I can do. Please let me know if there is a better way to solve this.
    - [x] Add `.close()` to unclosed file in `mediacloud/mediawords/db/handler.py`
    - [x] Add `.close()` to unclosed file in `mediacloud/mediawords/util/config.py`. These two involves modifying existing code in other files, hope you do not mind.

- [x] Remove the redundant code to serialize sentences to `JSON`. Sorry, I misunderstood you last time. You are right, this step is unnecessary, I have removed it now.

- [x] Rename `_eliminate_symbols()` to `_tokenize_sentence()`

- [x] Keep table names in the SQL queries. I was thinking the same as you, having them as parameters are hard to code as well.

- [x] Use the SQL query from the last PR review.
    - I wrote my SQL query in a nested structure because I was trying to use `LIMIT` and `OFFSET` to give better customization in the output (which might be useful if we want to focus on a small number of stories in the database or expand the database). 
    - My SQL query does what I want, although it might not be the most efficient way to do so. First, the code in the bracket selects a limit number of `stories_id`s (with `LIMIT` and `OFFSET`), then the outer code will only output a `stories_id, sentence` pair if this `stories_id` is in the `stories_id`s fetched by the inner code.
    - I did not use `ORDER BY story_sentences.sentence_number` because it does not matter whether the sentences are in the correct order or not, LDA only cares about the occurrence of each word.
    - I am now using your SQL query, but if we want to make use of `LIMIT` and `OFFSET`, we have to use a similar nested structure since there are two `ORDER BY` statements. I have also attached my SQL query in the code in case you would like to take a look. 
    - Please let me know if I missed or misunderstood anything. Much appreciated.

- [x] Write a short explanation in each of the three models in doc-strings.

PR #174, opened 2017-07-11
Added names of top public figures to user dict

natlungfy:
-The names are from [LIVAC's Annual Celebrity Roster (2014-16)](http://www.livac.org/celebrity.php?lang=en).
-Updated README to acknowledge sources.


pypt:
Natalie, could you please add a quick unit test (just a small `assert` or two in `test_zh.py`) that would ensure that the name of one of the celebrities gets tokenized correctly?

Say we choose to replace Jieba with some other Chinese language tokenization library in the future. In that case, it would be useful for us to learn that it is expected from this new library to be able to tokenize special cases such as celebrity names too.

natlungfy:
@pypt Added one unit test with a few names to tokenize. There was a test that also included one of the names in its sentence.

pypt:
Thanks, looks good, fighting with Travis now. The build is failing, but it seems that it's not due to your changes.

pypt:
Merged, deployed.

PR #199, opened 2017-10-03
Tidy defaults

ColCarroll:
This will at least make sure the imports in `mediawords.solr.run.solr` and `mediawords.solr.run.zookeeper` are tested.  I also removed `fqdn()` from being a default argument, since that evaluates at import time.


pypt:
Thanks Colin, looks good.

PR #201, opened 2017-10-10
Convert test suite to pytest

ColCarroll:
See #200 .  Got a little distracted in a few places and removed `*` imports, and fixed two `FIXME`s.  These are all changes to tests, not working code, so it is already tested!

### DIFF
--- a/ansible/roles/python-dependencies/vars/main.yml
+++ b/ansible/roles/python-dependencies/vars/main.yml
@@ -29,7 +29,7 @@ python_dependencies:
     - { name: "nltk", version: "3.2.4" }
 
     # Unit tests
-    - { name: "nose", version: "1.3.7" }
+    - { name: "pytest", version: "3.2.2" }
 
     # numerical processing, required by ForceAtlas2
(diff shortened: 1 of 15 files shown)


pypt:
Looks good to me.

Genuine question: what's wrong with `from module_that_is_being_tested import *` in unit tests?

Also, why in some cases you've changed it to `import module_that_is_being_tested as mc_module` and in some others you opted for `from module_that_is_being_tested import (function_1, function_2, ...)`?

ColCarroll:
`import *` is just an anti-pattern:  it makes it harder for automated linters, and causes strange bugs from a distance.  For example, in a few of those files, removing `import os` from `module_being_tested.py` would break the test suite, because it used the `os` module!

The two different styles of imports are just my own sense of what's readable -- a few of the test files have lots of imported functions, so namespacing them seemed like a good idea -- it is more explicit and more readable.  

For a file that's <40 lines, importing the function by name was less work on my part!

pypt:
Fair enough!

PR #203, opened 2017-10-10
Exit without exception in run_alone

### LINKED ISSUE #202: Inscrutable error after running `test_process.py`
Tests pass, but there's a funny error message afterwards.  In particular: 
```
$ py.test -v mediacloud/mediawords/util/test_process.py::test_run_alone
...
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/Users/colin/projects/mediacloud/mediacloud/mediawords/util/process.py", line 182, in __remove_run_alone_lock_file
    sys.exit(signum)
SystemExit
```

I may poke around and see if I can't figure that out, though it looks like it is a background task getting killed by `atexit`.

ColCarroll:
This fixes #202 , apparently `sys.exit` throws a `SystemExit` error (from [here](https://stackoverflow.com/questions/21705693/is-there-a-way-to-make-py-test-ignore-systemexit-raised-on-a-child-process), and [here](https://stackoverflow.com/questions/173278/is-there-a-way-to-prevent-a-systemexit-exception-raised-from-sys-exit-from-bei)).  This seems a little unsafe (I am not sure if it preempts other `atexit` functions, in particular), but tests keep passing cleanly.

### DIFF
--- a/mediacloud/mediawords/util/process.py
+++ b/mediacloud/mediawords/util/process.py
@@ -166,7 +166,9 @@ def run_alone(isolated_function: Callable, *args, **kwargs) -> Any:
         return unique_id
 
     # noinspection PyUnusedLocal
-    def __remove_run_alone_lock_file(signum: int = None, frame: int = None) -> None:
+    def __remove_run_alone_lock_file(signum: int = None,
+                                     frame: int = None,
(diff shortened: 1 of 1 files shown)


pypt:
Thank you!

PR #219, opened 2017-11-07
Add python requirements.txt

ColCarroll:
@pypt mentioned there might be a reason that we are using an ansible task to store python requirements instead of keeping it all in one place, but I figured I'd open a PR to talk about switching it over.  Some benefits of this are:

-  Python code is self contained
-  Better integration with the python ecosystem (virtual environments, linters, CI tools)
-  Easier to bootstrap developers into python codebase


pypt:
Remembered the (multi-level) reason, bear with me:

1. We use [Travis](https://travis-ci.org/berkmancenter/mediacloud) for CI testing. Travis is free for open-source projects (such as ours), is a pretty nice tool overall, and we don't need to maintain our own Jenkins, so that's cool.
2. Travis uses Ubuntu 14.04 (Trusty) for their instances. We run Media Cloud on Ubuntu 16.04 (Xenial). Before, we used to maintain special exceptions in our code just for being able to test our code on 14.04. Maintaining such exceptions was a major PITA (imagine all the `if [ $(lsb_release) == ... ]` in the code).
3. Travis supports running Docker instances, so to be able to test our code on 16.04, in [`.travis.yml`](https://github.com/berkmancenter/mediacloud/blob/master/.travis.yml#L88-L127), we pull Ubuntu 16.04, provision it using Ansible, copy the source tree to be tested, run the tests, copy test coverage results back to Travis's 14.04 instance and submit them to [Coveralls](https://coveralls.io/r/berkmancenter/mediacloud).
4. But then we have another problem - Docker containers make things way slower. You have to pull the container from [Docker Hub](https://hub.docker.com/r/dockermediacloud/) (1-2 minutes), provision an empty Ubuntu system with Ansible (10-15 minutes) and then run the tests, some of which (esp. the integration tests) are unfortunately quite slow as they start a bunch of services, submit something to those services and then expect something else back. Travis has a hardcoded 50 minute test run limit which we used to hit constantly with our monolithic code base. We've split up the tests since into four chunks so that all the tests manage to finish in 20-30 minutes in parallel, but it's still maddening to wait for half an hour to find out whether your code works on every `git pull`.
5. Naturally, a way to speed up the process is to pre-provision an empty Ubuntu 16.04 by installing the system-wide dependencies + Perl modules + Python modules, and then using the pre-provisioned container for Travis runs. That's what we do in [`ansible/Dockerfile`](https://github.com/berkmancenter/mediacloud/blob/master/ansible/Dockerfile) to shave off those 10 minutes or so out of every run (parallelized or not).
6. `ansible/Dockerfile` that does the pre-provisioning will naturally want to install Python dependencies somehow among other things. However, said `Dockerfile` had no way to access `../mediacloud/requirements.txt` because Docker's [`COPY` can't copy anything from parent directories](https://stackoverflow.com/a/24540011/200603) (kind of makes sense I guess). To get around this limitation, we could:
    1. `docker build -f ansible/Dockerfile .`, i.e. insist that the user must `cd` to Media Cloud's root, and then copy the whole 400 MB of Media Cloud's source code to the container that is being pre-provisioned just to run a bunch of scripts from `ansible/` -- IMHO confusing (one has to know from which directory Docker is supposed to be run; also, no way for `Dockerfile` to test whether it's being run from a correct directory), slow (more I/O done, possibly a bigger resulting container size), `ansible/` is no longer self-contained (system dependencies role depends on having the rest of the source code somewhere).
    2. Move `ansible/Dockerfile` to Media Cloud's root. Also confusing (one will then expect that said `Dockerfile` will install and run all the Media Cloud, not just pre-provision a container used for testing), slow (the whole source code has to be copied to the container), `ansible/` is still not self-contained (as it depends on `../mediacloud/requirements.txt`).
    3. Regretfully move all the Python dependencies from `requirements.txt` to a Ansible variable, just like [we do for Perl dependencies](https://github.com/berkmancenter/mediacloud/blob/master/ansible/roles/perl-dependencies/vars/main.yml#L17-L123). Confusing in some other way (*"where's the `requirements.txt`?"*), but maybe by not too much (*"we keep Python requirements in an Ansible variable, don't ask why, you don't wanna know"*), faster (only `ansible/` has to be copied to the container to do the provisioning), `ansible/` remains to be self-contained (everything that we need to install to run Media Cloud is under a single directory), probably a little bit more consistent (Perl and Python dependencies are defined in the exact same way). This seemed like the least bad option, so I went for it.
    4. Move `requirements.txt` to `ansible/roles/python-dependencies/files/`. Not that much different from having Python requirements in an Ansible variable IMHO, yet still confusing because the file is hard to find, so it's basically a question of whether it's easier to type `source ~/.virtualenvs/mediacloud/bin/activate; pip install -r ansible/roles/python-dependencies/files/requirements.txt` or `ansible-playbook --tags python-requirements setup.yml`.

If you were to say that we keep on digging on the rabbit hole, you'd be right. IMHO a real solution for strange constraints like this would be to split up our monolithic repository into multiple smaller ones which would then be independently tested, thus making the set up easier via simple "divide and conquer" (but then I'd expect us to have some other problems altogether :)). However, if you come up with a elegant solution for where to place `requirements.txt`, I'd be all ears. Maybe option four from the list above?

ColCarroll:
Thanks for the detailed answer -- that makes a lot of sense!  One question before I start suggesting more exotic fixes:  are you sure it hasn't fixed itself, either with travis getting better at caching or something else?  This build did not take significantly longer than any other one.

pypt:
The build itself doesn't rebuild the Docker container that we use as a testbed. It pulls the container with pre-installed system / Perl / Python dependencies and completes the provisioning by running the full Ansible playbook against it to get a complete picture. For example, if the pre-provisioned container already has `htop` APT package, the `apt install htop` command run by the Travis test build will essentially be a very fast no-op (because the package is already there).

What wouldn't work anymore is updating the pre-provisioned container (which one sometimes might want to do, e.g. to preinstall it with up-to-date packages so that subsequent `apt update && apt upgrade` becomes faster, preinstall new Perl / Python dependencies to speed things up again, etc.) In other words, [these commands used for recreating container](https://github.com/berkmancenter/mediacloud/blob/master/ansible/Dockerfile#L1-L12) will fail because Ansible won't be able to find `requirements.txt`.

pypt:
...so what your build did was:

1. Pull the Ubuntu Docker container pre-provisioned with most of Media Cloud's dependencies, the container itself created in [2017-09-28](https://github.com/berkmancenter/mediacloud/blob/master/.travis.yml#L19).
2. Copy the full source code that is being tested to said container (including your changes).
3. Run the Ansible playbook from that tested source code.
    * Given that it has a full source tree, Ansible was able to access `requirements.txt`.
    * Ansible playbook reused the dependencies already pre-installed on the test container from September and installed dependencies that were added to the requirements list since then.
4. With the fully provisioned environment, Travis run the tests on the Docker container which managed to pass.

pypt:
Option number 5 might be to move `requirements.txt` file to `ansible/roles/python-dependencies/files/` and then symlink `mediawords/requirements.txt` to `../ansible/roles/python-dependencies/files/requirements.txt`, Git kind of sort of supports symlinks, we have a few of them [in our Solr collection configuration](https://github.com/berkmancenter/mediacloud/tree/master/solr/collections/collection1). Not sure how ugly of a solution that would be.

ColCarroll:
I like the symlink idea!  

For what it is worth, a fresh install of the python dependencies on my machine takes ~55s, and travis [supports a pip cache](https://docs.travis-ci.com/user/caching/#pip-cache), which lowers the installation time on my machine to ~25s.

The next step in my evil plan is to hook circleci up to just test the python code, so there's a ~2 minute turnaround on python unittests, instead of ~20 mins.

pypt:
I'm not sure if another CI provider (e.g. CircleCI) would improve things by much -- you'd still have to install all those Perl / Python / system dependencies somehow and set up the test instance. Also, Circle doesn't seem to support 16.04, so you'll probably have to go the Docker way too, essentially recreating what we have now. I see those CI providers as pretty much identical in our case.

As for caching, the Docker Travis Media Cloud image serves as a single huge 800 MB cache for everything already, and it reaches the point where it starts to run unit tests (starting with Python ones) in ~7 minutes.

How about this for a plan regarding original intention of this PR (resurrecting `requirements.txt`):

- [ ] Sign up for [Docker Hub](https://hub.docker.com/), ping me with your username, I'll add you to the [`dockermediacloud` team](https://hub.docker.com/u/dockermediacloud/).
- [ ] Move `requirements.txt` to `ansible/roles/python-dependencies/files/`.
- [ ] Make Ansible task use `requirements.txt` from `files/`.
- [ ] In `mediacloud/`, create a symlink to `../ansible/roles/python-dependencies/files/requirements.txt`.
- [ ] Recreate [`mediacloud-travis` Docker container](https://github.com/berkmancenter/mediacloud/blob/master/ansible/Dockerfile#L1-L12), push it to Docker Hub, update container's tag in `.travis.yml` in your PR, test if it works.
- [ ] Merge & deploy & succeed in life.

What do you think?

ColCarroll:
The intention of migrating to CircleCI would be to isolate the python unittests, and rely on the existing (travis) tests for integration, and whatever OS level tests are need.  It would clone the repo, install the python dependencies, and run `py.test mediacloud/mediawords/`.  In particular, docker would not be involved.  

CircleCI also allows multiple different languages for a single repo, so I could imagine eventually moving the perl tests there too, but for now, it is just another CI service to give us automated testing and coverage for Python.  As far as I can tell, travis doesn't support that if we're already running one task there.

I'll get started with what you mentioned -- thanks for the plan!

pypt:
Thanks for the reply! Let's get `requirements.txt` back and continue CI discussion on a separate GH issue.

pypt:
Hey Carol, any progress on `ln -s`?

PR #221, opened 2017-11-10
Convert downloads to use requests instead of curl

ColCarroll:
Quick fix for these two FIXME's

### DIFF
--- a/mediacloud/mediawords/util/web/__init__.py
+++ b/mediacloud/mediawords/util/web/__init__.py
@@ -1,10 +1,11 @@
-from furl import furl
 import os
 import tempfile
 
+from furl import furl
+import requests
+
 from mediawords.util.log import create_logger
 from mediawords.util.perl import decode_object_from_bytes_if_needed
-from mediawords.util.process import run_command_in_foreground, McRunCommandInForegroundException
 
 log = create_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


pypt:
Thanks Carol! I've made this "rewrite to use native Python code" note to myself some time ago, but found out ever since that downloading big files using Python (requests, urllib2/3, ...) is something like 10x slower than just calling curl. We use `download_file()` to download Solr binaries (~150 MB), so using native code here would make it unbearably slow.

Closing for now, please reopen if you find a way to make it at least comparably as fast as curl. Bonus points for not using `requests` directly and going for `UserAgent` class instead.

ColCarroll:
Ah, if we're thinking about large files, this runs comparably fast to `cURL` on my machine:

```
def download(source_url, target_path):               
    r = requests.get(source_url, stream=True)        
    r.raise_for_status()                             
    with open(target_path, 'wb') as f:               
        for chunk in r.iter_content(chunk_size=1024):
            if chunk:                                
                f.write(chunk)                       
    return target_path                               
```

I can check and see if the user agent is amenable to a similar pattern.

pypt:
Tried it myself, `requests` is 2.5x slower (somehow it was 10x slower when I was downloading `GoogleNews-vectors-negative300.bin`), but we can live with that:

```
$ cat <<EOF > download_requests.py

import requests
import sys


def download(source_url, target_path):               
    r = requests.get(source_url, stream=True)        
    r.raise_for_status()                             
    with open(target_path, 'wb') as f:               
        for chunk in r.iter_content(chunk_size=1024):
            if chunk:                                
                f.write(chunk)                       
    return target_path


download(source_url=sys.argv[1], target_path=sys.argv[2])

EOF

$ time curl --silent --output /var/tmp/solr-curl.tgz http://www-eu.apache.org/dist/lucene/solr/7.1.0/solr-7.1.0.tgz
curl --silent --output /var/tmp/solr-curl.tgz   0.10s user 0.48s system 28% cpu 2.047 total

$ time python3.5 download_requests.py http://www-eu.apache.org/dist/lucene/solr/7.1.0/solr-7.1.0.tgz /var/tmp/solr-requests.tgz
python3.5 download_requests.py  /var/tmp/solr-requests.tgz  3.94s user 0.69s system 91% cpu 5.061 total
```

 It still would be nice to reuse `UserAgent` somehow for downloading (possibly large) files, at the very least because:

* `requests` has no default timeout
* `requests` doesn't retry on HTTP errors by default
* `UserAgent` sets correct `User-Agent`
* general niceness of having a single class for dealing with the web stuff

`UserAgent` has [`get()` method](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/mediawords/util/web/user_agent.py#L235-L252) which can fetch stuff but [stores it in a variable](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/mediawords/util/web/user_agent.py#L771-L782) which is not cool when downloading large files.

I guess the best approach would be to move `download_file()` and its buddy to `UserAgent`, move [`requests` request preparation](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/mediawords/util/web/user_agent.py#L677-L685) and [request sending](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/mediawords/util/web/user_agent.py#L692-L699) to helpers (so we can reuse them everywhere), and implement the data buffering to a file in disk as you did in your sample code. What do you think?

ColCarroll:
Sorry for leaving this for so long! It looks like `UserAgent` is a useful class, but I have trouble figuring out what is going on where with it, so I cannot be much help moving any of it around.

pypt:
That's unfortunate, it could only mean that I've done insufficient job with naming and documentation in `UserAgent` :(. Could you help me to improve the readability of the class then?

Would you be able to clarify which parts you don't understand? Have you had a look at [unit tests](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/mediawords/util/web/test_user_agent.py#L47-L72)?

hroberts:
closing.  we can update to useragent in the future if necessary.

PR #222, opened 2017-11-10
Feed finder

ColCarroll:
Still needs documentation, examples, and tests, but this is a working python port of the feed extractor.  Uses a combination of logic from the perl implementation and from https://github.com/dfm/feedfinder2 (which is a python3 port of http://www.aaronsw.com/2002/feedfinder/)

The big design change is that `generate_feed_urls` is a generator, and returns urls in order of likelihood (so better heuristics go first).  It can take quite a long time to iterate through all of them, but this gives the control to grab one or two instead.  There is also `find_feed_url` which just gives you the first one.

### DIFF
--- a/mediacloud/mediawords/feed_finder/feed_finder.py
+++ b/mediacloud/mediawords/feed_finder/feed_finder.py
@@ -1,0 +1,160 @@
+from urllib.parse import urljoin
+
+from bs4 import BeautifulSoup
+
+from mediawords.util.web.user_agent import UserAgent, McUserAgentException
+
+
+def _is_feed_url(url):
+    endings = (
+        '.rss',
+        '.rdf',
+        '.atom',
+        '.xml',
+    )
+    url_lower = url.lower()
+    return any(url_lower.endswith(ending) for ending in endings)
+
+
+def _might_be_feed_url(url):
+    substrings = (
+        'rss',
+        'rdf',
+        'atom',
+        'xml',
+        'feed'
+    )
+    url_lower = url.lower()
+    return any(substring in url_lower for substring in substrings)
+
+
+class FeedFinder(object):
+
+    def __init__(self, url, html=None):
+        self.url = url
+        self._html = html
+        self._soup = None
+
+    @property
+    def html(self):
+        """String of the html of the underlying site."""
+        if self._html is None:
+            ua = UserAgent()
+            try:
+                response = ua.get_follow_http_html_redirects(self.url)
+            except McUserAgentException:
+                self._html = ''
+            else:
+                self._html = response.as_string()
+        return self._html
+
+    @property
+    def soup(self):
+        """BeautifulSoup representation of the data."""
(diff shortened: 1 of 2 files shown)


ColCarroll:
This is documented and tested -- let me know if anything else should be added.

pypt:
Didn't do in-depth review yet, just skimmed through:

* You seem to using fstrings which are a feature from Python 3.6. As of today, we're using 3.5: https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#use-python-35
* No type hints: https://github.com/berkmancenter/mediacloud/blob/master/doc/coding_guidelines.markdown#declare-function-parameter-and-return-value-types
* Some methods are probably designed to be protected / private and should have appropriate prefixes (`_` or `__`): https://stackoverflow.com/a/1301369
* Have you had a look at the unit test of the current feed scraper (https://github.com/berkmancenter/mediacloud/blob/master/lib/MediaWords/Feed/Scrape/t/feed_scrape.t)? Is your feed finder able to parse out feeds from all the test cases from said test? If so, would you consider porting the unit test too? Your current test seems to be trying out only a couple of basic cases (`<link />` and some hardcoded paths, e.g. `/index.xml`)
* No stats for how this feed finder is performing compared to the current implementation as noted in https://github.com/berkmancenter/mediacloud/issues/227#issuecomment-351476234

pypt:
> I am pretty sure there are no f-strings here

My bad, there aren't.

> Are type hints actually a requirement to merge? The project does not use them anywhere.

They are being used pretty much everywhere, e.g. see [`def encode_json(json_obj: Union[Dict, List], pretty: bool = False) -> str`](https://github.com/berkmancenter/mediacloud/blob/master/mediacloud/mediawords/util/json.py#L25).

> I do not intend any of the methods or properties to be marked as private.

Why not?

ColCarroll:
Hey, so this is in a separate repo now – I have not released it on PyPI since I want it to have a stable-er home than my user (I do not have the authority to push anywhere else!). See https://github.com/ColCarroll/feed_seeker

@hroberts just made a change that adds both the spidering you mention.  It does quite a bit better than the non-spidering version (at the expense of speed!), but since it uses a generator, it is easy to let things timeout.  For example, running on [this url](https://www.nytimes.com/2018/01/06/opinion/looming-digital-meltdown.html?rref=collection%2Fsectioncollection%2Fopinion&action=click&contentCollection=opinion&region=rank&module=package&version=highlights&contentPlacement=4&pgtype=sectionfront) found no feeds, but can find feeds with a spider. On my machine it is 13 seconds to find the "Sunday Opinion" feed, and 28 seconds to also find the main "Opinions" rss and a "Opinions Letters" rss. 

With respect to both of your comments on testing this on "live evaluation data", I am not sure what that means. If someone could run the perl script against a set of urls, I would be happy to run the Python script, and we could compare the outputs, which might help.

Issue #224, opened 2017-11-13
bad urls from topic spider are crashing python web fetching code


hroberts:
Bad urls from the topic spider are crashing the new python web fetching code in various places.  The latest is in the error message for topic 1848.

rahulbot:
"
[...]

PR #225, opened 2017-11-13
Validate urls before scraping

ColCarroll:
I am pattern matching a bit here, and the git commit hook (`perltidy`) failed, but then it failed when I removed my changes, so... would love a perl-literate review.

This should help #224 in the future, though I think there still needs to be a change to handle errors when they happen there.


pypt:
Thanks Colin, looks like a reasonable fix.

PR #238, opened 2017-12-06
Fix python test suite

ColCarroll:
Python tests do not pass currently -- `test_s3_credentials` is discovered by the test runner, and will be `None` if they are not available.  This also exposed a (now fixed) bug in the actual function.

Also marks `test_fqdn` as an expected failure, since it does not pass locally.

### DIFF
--- a/mediacloud/mediawords/key_value_store/test_amazon_s3.py
+++ b/mediacloud/mediawords/key_value_store/test_amazon_s3.py
@@ -1,10 +1,11 @@
 import pytest
 
 from mediawords.key_value_store.amazon_s3 import AmazonS3Store
-from mediawords.key_value_store.test_amazon_s3_credentials import TestAmazonS3CredentialsTestCase, test_s3_credentials
+from mediawords.key_value_store.test_amazon_s3_credentials import (
+    TestAmazonS3CredentialsTestCase, get_test_s3_credentials)
 
 
-test_credentials = test_s3_credentials()
+test_credentials = get_test_s3_credentials()
 
 pytest_amazon_s3_credentials_set = pytest.mark.skipif(
     test_credentials is None,
--- a/mediacloud/mediawords/key_value_store/test_amazon_s3_credentials.py
+++ b/mediacloud/mediawords/key_value_store/test_amazon_s3_credentials.py
@@ -10,7 +10,7 @@ from mediawords.util.config import get_config as py_get_config
 from mediawords.util.text import random_string
 
 
-def test_s3_credentials() -> Union[dict, None]:
+def get_test_s3_credentials() -> Union[dict, None]:
     """Return test Amazon S3 credentials as a dictionary or None if credentials are not configured."""
 
     config = py_get_config()
@@ -31,12 +31,13 @@ def test_s3_credentials() -> Union[dict, None]:
         credentials = copy.deepcopy(config['amazon_s3']['test'])
 
     # We want to be able to run S3 tests in parallel
-    credentials['directory_name'] = credentials['directory_name'] + '-' + random_string(64)
+    if credentials is not None:
+        credentials['directory_name'] = credentials['directory_name'] + '-' + random_string(64)
 
     return credentials
 
 
-test_credentials = test_s3_credentials()
+test_credentials = get_test_s3_credentials()
 
 pytest_amazon_s3_credentials_set = pytest.mark.skipif(
     test_credentials is None,
--- a/mediacloud/mediawords/key_value_store/test_cached_amazon_s3.py
+++ b/mediacloud/mediawords/key_value_store/test_cached_amazon_s3.py
@@ -1,9 +1,10 @@
 import pytest
 
 from mediawords.key_value_store.cached_amazon_s3 import CachedAmazonS3Store
-from mediawords.key_value_store.test_amazon_s3_credentials import TestAmazonS3CredentialsTestCase, test_s3_credentials
+from mediawords.key_value_store.test_amazon_s3_credentials import (
+    TestAmazonS3CredentialsTestCase, get_test_s3_credentials)
 
-test_credentials = test_s3_credentials()
+test_credentials = get_test_s3_credentials()
 
 pytest_amazon_s3_credentials_set = pytest.mark.skipif(
     test_credentials is None,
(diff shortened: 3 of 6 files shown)


pypt:
Thanks.

`fqdn()` test fails for you because either (as per the implementation):

* Your hostname is not set
* Your hostname includes underscore (`_`)
* Your hostname doesn't resolve

We want this test because without a valid (resolving, non-underscore-including) FQDN you will get very weird results when running Solr (locally or in production), especially with shards.

Could you revert the FQDN bit?

ColCarroll:
> We want this test because without a valid (resolving, non-underscore-including) FQDN you will get very weird results when running Solr (locally or in production), especially with shards.

I can revert, but this is an antipattern to test the underlying environment in a unittest.

pypt:
How would you suggest us to test it?

ColCarroll:
Just pushed some changes that isolate the functionality of `fqdn` a bit more.  The tests for this file seem to be just testing the `socket` library, which does not seem great (especially since they also require an internet connection).

pypt:
Colin, if your machine accidentally had a hostname with an underscore (`colin_carroll`) or some other one which doesn't resolve, or no hostname at all (which happens with EC2 instances using some AMIs), some of the tests (belonging to the group of "integration" tests I guess because they don't test code "units" as such but rather how and whether various components work together) might suddenly stop working for mysterious reasons.

Our current integration tests aren't that friendly because they operate similarly to the Deep Thought from the The Hitchhiker's Guide to the Galaxy. In the book, Deep Thought, the biggest, coolest computer in the universe, is asked to come up with an answer to the ultimate question of life, the universe, and everything. Deep Thought takes a while to think about it (7,5 million years specifically) and then spits out the answer: 42. In a similar fashion, our integration tests take a while to run and just spit out a result at the end without being overly specific of what has failed; one could remedy this with better unit (not integration) test coverage in various parts, but we're still working on that.

If you were to dig in and try to find out the reason of the sudden test failure, after 30 minutes of debugging prints everywhere you'll find out that Solr is refusing to start (those nasty integration tests start and stop Solr for their own purposes). After another 30-60 minutes, you'll be forced to learn about the niceties of Solr sharding and that Solr is very specific about how it likes their hostnames: it has to be FQDN, it has to be resolveable, and it can't have underscores (which Solr would translate to `/` and then try to resolve everything that comes before the slash).

`fqdn()` is used solely by Solr exec scripts to resolve, verify and pass the fully qualified hostname for Solr to use. The rest of the system doesn't care about what's the name of the host it runs on. I totally understand what you're trying to do, yes, we shouldn't rely on the environment, but tell that to Solr. I've personally spent ~4 hours figuring out the Solr hostname issues, and this small test is my knowledge on the topic encoded in the form of code. `test_fqdn()` is definitely not a unit test (given that it tests `socket` and whether the hostname is set to the reasonable value), I'm not even quite sure which bucket of testing from the pages of *Code Complete 2* does it belong to, but it does its little job of preventing Solr shards from going insane.

Lastly, there are at least three ways to get FQDN: `socket.getfqdn()` (doesn't work on all machines), `subprocess.check_output(["hostname", "-f"])` (not that great as it's a system command), and `socket.getaddrinfo(socket.gethostname(), 0, flags=socket.AI_CANONNAME)[0][3]` (current implementation). How would you suggest that we ensure (with whichever type of test) that we've chosen the right way to get the FQDN with parameterized `hostname=None` as per your commit?

So, can we leave it as-is for the time being?

ColCarroll:
> How would you suggest that we ensure (with whichever type of test) that we've chosen the right way to get the FQDN with parameterized hostname=None as per your commit?

Hmm... I am not sure I understand.  Are you asking how to use this new version of `fqdn`?  In that case, nothing needs to change.  If you call `fqdn()`, it has the same behavior as before this commit.

Are you asking how we can make sure that the tests fail if the machine has an underscore in its domain name?  In that case I would argue that the unittest suite is a really bad place to be putting runtime assertions. 

In particular, the Python test suite was broken by the key-value store changes, but those bugs were not picked up, in part because testing was not done in an isolated environment.  See here [for](https://circleci.com/gh/ColCarroll/mediacloud/17) circleci integration ready to be turned on when we have a test suite that can run.

pypt:
In short, how do we verify that getting FQDN via chosen method (`socket.getfqdn()`, `subprocess.check_output(["hostname", "-f"])`, or `socket.getaddrinfo(socket.gethostname(), 0, flags=socket.AI_CANONNAME)[0][3]` works if we pass a test string to `fqdn()` avoiding the underlying syscall altogether?

Also, where would you suggest we put the assertion for underscores and such?

Offtopic, but your Circle config is failing because: 1) you have no `mediawords.yml`; 2) you have no test S3 credentials set (we set them in [environment variables](https://github.com/berkmancenter/mediacloud/blob/master/.travis.yml#L36-L45)).

ColCarroll:
> In short, how do we verify that getting FQDN via chosen method (socket.getfqdn(), subprocess.check_output(["hostname", "-f"]), or socket.getaddrinfo(socket.gethostname(), 0, flags=socket.AI_CANONNAME)[0][3] works if we pass a test string to fqdn() avoiding the underlying syscall altogether?

We should not write unittests for the standard library or underlying OS

> Also, where would you suggest we put the assertion for underscores and such?

Wherever the underscores matter!  ```if not fqdn_is_ok_for_solr(host_tqdm): raise RuntimeError('descriptive error')```

> Offtopic, but your Circle config is failing because: 1) you have no mediawords.yml; 2) you have no test S3 credentials set (we set them in environment variables).

Adding credentials would definitely fix it, but also this PR fixes the underlying problem.

ColCarroll:
Closing per private discussion.

pypt:
Why close? d44df878671cc30d685133ff7ac7cf8d541bac14 is a good change, except for the FQDN part. Could you revert that bit and then we could then merge?

ColCarroll:
My mistake -- you had said that my tests were failing because "1) you have no mediawords.yml; 2) you have no test S3 credentials set (we set them in environment variables)."  I thought this meant you aren't worried about tests passing locally.

pypt:
Could you remove the `@pytest.mark.xfail` part too? In `test_network.py`:

https://github.com/berkmancenter/mediacloud/pull/238/files#diff-04184f358eda0760dc480d9c42802428

ColCarroll:
Would rather close - at that point this is just renaming some functions.

For what it is worth, I still think it is a bad idea to have a test suite that can only run on a specific network, and this PR had two different ways of fixing that.

PR #242, opened 2017-12-14
On "requests" exceptions, set response's body to the exception message

pypt:
As per LWP::UserAgent's (from which we inherit our UserAgent class
design) philosophy, if the parameters are right, UserAgent should never
raise and instead always return a valid Response object. To see whether
the request has failed, one is expected to call is_success() method.

On "requests" failures (when "requests" crashes being unable to parse
something), we create a Response object for with the exception message
in its body.

Before, this was being done incorrectly. In this commit, we set the body
correctly (in a way that can later be read by response.raw.stream()) and
also test whether the exception message got returned in HTTP body.

### DIFF
--- a/mediacloud/mediawords/util/web/user_agent/__init__.py
+++ b/mediacloud/mediawords/util/web/user_agent/__init__.py
@@ -16,7 +16,7 @@ import requests
 from furl import furl
 from requests.adapters import HTTPAdapter
 from requests.auth import HTTPBasicAuth
-from urllib3 import Retry
+from urllib3 import Retry, HTTPResponse
 
 from mediawords.util.config import get_config as py_get_config
 from mediawords.util.log import create_logger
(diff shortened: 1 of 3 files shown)


mediacloud/mediawords/util/web/user_agent/__init__.py
```diff
@@ -723,7 +733,7 @@ def __execute_request(self, requests_prepared_request: requests.PreparedRequest)
 
             requests_response.history = []
```
ColCarroll: `requests.TooManyRedirects` I guess returns a valid response?

pypt: Yes, it returns the last fetched page. Perl's `LWP::UserAgent` wouldn't report much of an error and just return the last fetched page, so we kind of do it that here too.

PR #243, opened 2017-12-14
Test normalized URL in is_http_url()

### LINKED ISSUE #224: bad urls from topic spider are crashing python web fetching code

pypt:
is_http_url() is a helper that we use for making sure that a string
looks like a valid URL before attempting to fetch it.

One could say that this is inexact science as the only way to find out
whether we can fetch the URL is to try to actually fetch it and catch
exceptions and whatnot. However, due to the quite big feature set of the UserAgent
class (max. size, max. redirects, HTTP / HTML redirects, backoff
retries, ...), it would be risky to add a single try-catch block around
every call to "requests" because that could hide potential programming
errors that we've made in this very young class by porting it to Python.

So, callers to UserAgent class are expected to call is_http_url() first
for their string arguments to make sure that the parameter is an URL
(otherwise one of the UserAgent's methods will do that for them and
raise an exception).

Before, is_http_url() was using furl as one of the methods of testing
the URL; however, it wasn't verifying all of its accessors, i.e. it
wasn't testing whether an URL can be constructed back from a furl
object. Because of that, is_http_url() was reporting the following URL
as valid:

http://michigan-state-football-sexual-assault-charges-arrest-players-names

...but then "requests" would normalize the URL by adding a slash to it
and subsequent calls to is_http_url() would fail for URL:

http://michigan-state-football-sexual-assault-charges-arrest-players-names/

In this commit (PR), is_http_url() stringifies parsed furl object back
to an URL and also tries to validate a normalized version of the URL.

(Hopefully) fixes #224.

### DIFF
--- a/mediacloud/mediawords/util/url/__init__.py
+++ b/mediacloud/mediawords/util/url/__init__.py
@@ -90,8 +90,18 @@ def is_http_url(url: str) -> bool:
     try:
         uri = furl(url)
 
-        # Some URLs have invalid paths that furl's constructor doesn't check
-        _ = str(uri.path)
+        # Try stringifying URL back from the furl() object to try out all of its accessors
+        _ = str(furl)
+
+        # Some URLs become invalid when normalized (which is what "requests" will do), e.g.:
+        #
+        #     http://michigan-state-football-sexual-assault-charges-arrest-players-names -- valid
+        #     http://michigan-state-football-sexual-assault-charges-arrest-players-names/ -- invalid (decoding error)
+        #
+        # ...so try the same with normalized URL
+        normalized_url = url_normalize.url_normalize(url)
+        normalized_uri = furl(normalized_url)
+        _ = str(normalized_uri)
 
     except Exception as ex:
         log.debug("Cannot parse URL: %s" % str(ex))
--- a/mediacloud/mediawords/util/url/__init__.py
+++ b/mediacloud/mediawords/util/url/__init__.py
@@ -91,7 +91,7 @@ def is_http_url(url: str) -> bool:
         uri = furl(url)
 
         # Try stringifying URL back from the furl() object to try out all of its accessors
-        _ = str(furl)
+        _ = str(uri)
 
         # Some URLs become invalid when normalized (which is what "requests" will do), e.g.:
         #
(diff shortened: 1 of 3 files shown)


mediacloud/mediawords/util/url/__init__.py
```diff
@@ -90,8 +90,18 @@ def is_http_url(url: str) -> bool:
     try:
         uri = furl(url)
 
-        # Some URLs have invalid paths that furl's constructor doesn't check
-        _ = str(uri.path)
+        # Try stringifying URL back from the furl() object to try out all of its accessors
+        _ = str(furl)
```
ColCarroll: should this still be `uri.path`?

mediacloud/mediawords/util/web/test_user_agent.py
```diff
@@ -1032,6 +1037,14 @@ def test_get_http_auth(self):
 
         hs.stop()
 
+    def test_get_request_invalid_url(self):
+        """request() with invalid URL."""
+        with pytest.raises(McUserAgentRequestException):
+            Request(method='GET',
+
```
ColCarroll: inadvertent space? also, the comment can be passed into `pytest.raises`: 

```
with pytest.raises(McUserAgentRequestException, message='Expected "Invalid IDNA error"'):
...
```

mediacloud/mediawords/util/url/__init__.py
pypt: Thanks, fixed, it should have been `str(uri)`.

Stringifying the whole furl object covers more base, e.g. try:

```python
import furl

# Invalid URL which gets treated as IDNA for whatever reason
x = furl.furl('http://michigan-state-football-sexual-assault-charges-arrest-players-names')

# Prints out empty string -- bad, we want it to fail to learn that the URL is bogus
print(x.path)

# Raises -- good!
print(x)
```

mediacloud/mediawords/util/web/test_user_agent.py
pypt: Thanks again, updated to just a comment (as `message` is not much used elsewhere).

ColCarroll:
Any idea on the travis failure?

pypt:
No idea. I've removed a chatty log message which was presumably filling up the Travis log, and it passes now:

https://travis-ci.org/berkmancenter/mediacloud/builds/316996904

Issue #215, opened 2017-11-02
generate and store topic-snapshot word2vec models


rahulbot:
We need to implement a job to generate word2vec models and store them in the database (as discussed in https://github.com/berkmancenter/mediacloud-systems/issues/54).

Per our discussion, this involves:

- [ ] a Python worker that accepts `topics_id` and `snapshots_id` and generates a model ([here's beta code that does it](https://github.com/mitmedialab/MediaCloud-WordEmbeddingsServer/blob/master/scripts/create-topic-model.py))
- [ ] a hook so that the worker is kicked off whenever a new spider is finished
- [ ] a new database table to hold foreign keys to the topic/snapshot, a (`BYTEA`?) blob that is the serialized model, and creation date
- [ ] an API method to retrive models created after a certain date so our WordEmbeddings server and query the API to fetch any models it doesn't have yet and save them locally (with an optional argument that limit it to a certain topic, and another that limits it to the latest snapshot for topics only)

rahulbot:
Note: we are now using a `vocab_size = 5000` param, and these models are much smaller and only one file (ie. from just a few kilobytes to 10s of megabytes)

pypt:
Depends on:

* [ ] #229

pypt:
Moved to milestone 6 because milestone 5 is due by 20 days.

pypt:
Some questions (@rahulbot?):

* So we want to remove stop words from sentences that we feed to the word2vec model? Internet forums can't seem to decide on the right answer, but I guess that depends on what exactly we want to do with the model?
* Same with stemming - would you like words to be stemmed before they get into the model?

rahulbot:
Nope on both. Our current code sends the full story text as is into `gensim.models.Word2Vec` as an array of sentences, all lowercase.  Very simple - available [to review in our prototype server](https://github.com/mitmedialab/MediaCloud-WordEmbeddingsServer/blob/master/scripts/create-topic-model.py#L62).

pypt:
Out of the curiosity, how does the model then deal with `Obama` / `Obamas` / `Obama's` and similar differences in word forms?

rahulbot:
Good question! I think it doesn't.  Each gets their own vector in the space the model creates as we've coded it right now.

The question is whether in this context of use we want to look at stem usage or not.  Right now these models are intended to drive:
1. visualizations of 2d spatially laid-out word clouds for topics
2. suggestions for related words while filtering a topic by a search term or creating a subtopic

[...]

PR #252, opened 2017-12-30
Rewrite language support classes to Python

### LINKED ISSUE #251: Rewrite language modules to Python
[`MediaWords::Languages::*`](https://github.com/berkmancenter/mediacloud/tree/master/lib/MediaWords/Languages) encapsulates natural language support code (stemming, text -> sentence and sentence -> word tokenization, etc.)

We need this code rewritten to Python to accommodate #215 (because splitting sentences into words by `\s` as is done in the beta example wouldn't cut it).

pypt:
As a dependency to #215 (and as a part of a general Python rewrite effort), I've rewritten our language support modules to Python.

Fixes #251. Travis build is currently failing due to Superglue live feed failure.

## Design

Similar to the current Perl implementation, the language support hierarchy consists of:

* [Abstract class `AbstractLanguage`](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/__init__.py#L46) which declares the methods to be implemented by concrete language classes, e.g. `stop_words_map(self) -> Dict[str, bool]` method to return a `dict` / hashref of stop words;
* Multiple helper classes to be inherited from by the concrete language implementations, based on the [mix-in pattern](https://en.wikipedia.org/wiki/Mixin#In_Python) widely used in Python. The mix-in classes are supposed to "help" the concrete language implementations by providing various `AbstractLanguage` methods:
    * [`SpaceSeparatedWordsMixIn` class](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/__init__.py#L94) - implements `split_sentence_to_words()` for languages for which words are separated by spaces.
    * [`SentenceSplitterMixIn` class](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/__init__.py#L126) - implements `split_text_to_sentences()` for languages that are supported by [`sentence_splitter` module](https://pypi.python.org/pypi/sentence_splitter) (more on the module later).
    * [`PyStemmerMixIn` class](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/__init__.py#L217) - implements `stem_words()` for languages that are supported by [PyStemmer module](https://pypi.python.org/pypi/PyStemmer/).
    * [`StopWordsFromFileMixIn` class](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/__init__.py#L264) - implements `stop_words_map()` for languages for which the stop word list is to be read from an external text file.
* Concrete language support implementations, e.g. see [`EnglishLanguage` class](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/en/__init__.py). A typical Latin-derived language implementation doesn't look like much, it merely implements `language_code()`, subclasses `AbstractLanguage`, all the mix-ins, and is good to go.
* [Factory class `LanguageFactory`](https://github.com/berkmancenter/mediacloud/blob/language_modules_python/mediacloud/mediawords/languages/factory.py#L75) which keeps a list of supported + enabled languages and creates + returns language instances on demand, based on their language code.

## Notes

* Words are being stemmed using [PyStemmer](https://pypi.python.org/pypi/PyStemmer/) which supports the same language set as Perl's [`Lingua::Stem::Snowball`](https://metacpan.org/pod/Lingua::Stem::Snowball).
* PyStemmer doesn't have stemmers for Catalan and Lithuanian languages, so I've added Snowball source files as Git submodules and generated Python stemmer classes out of them that get used by the respective language modules (4821bd8 for Catalan, cf47932 for Lithuanian). Generating C Snowball stemmers would make stemming slightly faster, but we have a minuscule amount of Catalan / Lithuanian language stories, and maintenance of those separate CPython modules would be a medium-to-major hassle, so I didn't bother. Autogenerated Python stemmer classes use some base classes from the Snowball package, so Snowball itself is added as a Git submodule as well (cd7c178); the classes are then referenced via a symlink.
* I really wanted to use NLTK's [`PunktSentenceTokenizer`](http://www.nltk.org/api/nltk.tokenize.html#nltk.tokenize.punkt.PunktSentenceTokenizer) for splitting text into sentences but couldn't because: 1) it doesn't have tokenization models for every language that we support -- Catalan, Hausa, Hindi, Hungarian, Lithuanian, Romanian, Russian models are missing, and to create those we'd have to get a corpus from somewhere for each of those languages and train our own models; and 2) it's not that great because out of the box it hasn't been taught to recognize many abbreviations (e.g. `sent_tokenize("Sgt. Pepper's Lonely Hearts Club Band")` would get split into `['Sgt.', "Pepper's Lonely Hearts Club Band"]`). So, I went ahead and rewrote Perl's `Lingua::Sentence` heuristic code to Python, split it up into a [separate repository](https://github.com/berkmancenter/mediacloud-sentence-splitter), released it as a [Python module](https://pypi.python.org/pypi/sentence_splitter), and made our code use it.

## Also

* Renamed `get_sentences()` to `split_text_to_sentences()` (8b91ce0), `tokenize()` to `split_sentence_to_words()` (2560f98). In academic literature that I've read five years ago, authors use the term "tokenization" for both text -> sentence and sentence -> word splitting, so I've tried to make the difference more explicit.
* Renamed `get_stop_words()` to `stop_words_map()` (78d0bcf) to suggest that the method returns a map (`Dict[str, bool]` in Python, hashref with stop word keys and `1`s as values in Perl) instead of a list of stop words.
* Renamed `get_language_code()` to `language_code()` (4cf6e28) for the sake of consistency (no other methods have `get_` prefix).
* Renamed `stem()` to `stem_words()` (872ffeb) to suggest that the method expects a `list` / arrayref parameter with multiple words to be stemmed instead of just a single word.


pypt:
Merged, deployed, seems to work.

After pulling, don't forget to:

```
# Refetch submodules as they've been moved around and some new ones have been added too
git submodule update --init

# Install some new Python modules
./script/run_in_env.sh pip install -r mediacloud/requirements.txt

# Remove obsolete directory
rm -rf lib/MediaWords/Languages/resources/
```

PR #254, opened 2018-01-03
Remove __str__ / __repr__ methods in Request / Response objects

### LINKED ISSUE #245: Topic mining fails with UnicodeEncodeError
```
                                                          message                                                          
---------------------------------------------------------------------------------------------------------------------------
 UnicodeEncodeError: 'utf-8' codec can't encode character '\udf33' in position 26352: surrogates not allowed              +
                                                                                                                          +
 During handling of the above exception, another exception occurred:                                                      +
                                                                                                                          +
 Traceback (most recent call last):                                                                                       +
   File "/home/mediacloud/mediacloud/mediacloud/mediawords/util/web/user_agent/response/response.py", line 68, in __repr__+
     for key, value in sorted(self.headers().items()):                                                                    +
 SystemError: PyEval_EvalFrameEx returned a result with an error set                                                      +
```

pypt:
I only halfway know what I'm doing here, but it appears that:

1. `UserAgent`'s `Request` and `Response` objects have `__str__()` implemented which returns human-readable string representation of the object
2. Both classes have Perl-ish `as_string()` method which is an alias for `__str__()`. `as_string()` was mostly used for debugging reasons.
3. Some Python code (I wasn't able to identify which, but probably not ours) was stringifying various `Response` objects for unknown reasons and printing them to STDOUT.
4. Python's `encode()` (or, in this case, `decode()?` I'm not sure) seems to be able to handle Unicode surrogate characters (e.g. `\ud800`) but raises `UnicodeEncodeError` when trying to print them to STDOUT.
5. So, as a result I've removed `__str__()` and `as_string()` methods so that no surrogate characters are attempted at being printed to STDOUT (or at least now it would be easier to find out which of our own code - if any - attempts to print said characters).

(Hopefully) fixes #245.

No unit tests because no new code, also not quite sure what to test (as I honestly don't know which code calls `__str__()`).

Not requesting a review for this comparably small and experimental fix.

### DIFF
--- a/lib/MediaWords/Controller/Admin/Feeds.pm
+++ b/lib/MediaWords/Controller/Admin/Feeds.pm
@@ -170,7 +170,7 @@ sub invalid_syndicated_feed
     my $ua       = MediaWords::Util::Web::UserAgent->new();
     my $response = $ua->get( $feed->{ url } );
 
-    return "url fetch for '$feed->{ url }' failed: " . $response->as_string if ( !$response->is_success );
+    return "url fetch for '$feed->{ url }' failed: " . $response->decoded_content if ( !$response->is_success );
 
     eval { XML::FeedPP->new( $response->decoded_content ) };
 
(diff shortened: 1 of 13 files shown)


pypt:
Oh, and CI is still failing due to Superglue downtime.

PR #255, opened 2018-01-05
Rewrite color palette generator code to Python

pypt:
Some web UI code needs to generate matching color palettes for coloring various partisanships and whatnot.

I rewrote color palette generator to Python. Should work identically to the Perl code.

I've tried to find an analogous color palette generator to Python but failed so I rewrote some of the [`Color::Mix`](https://metacpan.org/pod/Color::Mix) helpers to Python too.

### DIFF
--- a/mediacloud/mediawords/util/colors.py
+++ b/mediacloud/mediawords/util/colors.py
@@ -26,20 +26,22 @@ def rgb_to_hex(r: int, g: int, b: int) -> str:
     return '%02x%02x%02x' % (int(r), int(g), int(b),)
 
 
-def analogous_color(color: str, my_slices: int = 3, slices: int = 32) -> List[str]:
-    """Analogous color schemes are ones in which the colors lie next to each other on the color wheel.
+def analogous_color(color: str, return_slices: int = 4, split_slices: int = 12) -> List[str]:
+    """Generate analogous color scheme starting with the provided color.
 
-    By default this method splits up the colorwheel into 12 pieces and returns the next 3 pieces of the wheel. So using
-    this if you entered 0000ff (blue) then by default you would get back '80000ff' (purple), 'ff00ff' (pink) and
-    'ff0080' (hotpink) colors.
+    Analogous color scheme is the one in which the colors lie next to each other on the color wheel.
 
-    If you split the wheel up into 36 as in the example above you essentially reduce the range of the colors by 3. So
-    now if you entered 0000ff (blue) but added a parameter 3 (For number of colors you want returned) and 36 (To split
-    the wheel up into 36 pieces) then you would get back '2b00ff' (Blue), '5500ff' (Bluey Purple) and '8000ff' (Purple).
+    By default this method splits up the color wheel into 12 pieces and returns the original parameter color plus the
+    next 3 pieces of the wheel (4 colors in total). For example, by padding '0000ff' (blue) you would get back '0000ff'
+    (blue -- the original color) '80000ff' (purple), 'ff00ff' (pink) and 'ff0080' (hot pink) colors.
 
-    This gives and easy way to either limit or expand the range of the analogous color palette.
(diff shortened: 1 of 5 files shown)


mediacloud/mediawords/util/test_colors.py
```diff
@@ -0,0 +1,86 @@
...
+def test_rgb_to_hex():
+    assert rgb_to_hex(255, 0, 0).lower() == 'ff0000'
+    assert rgb_to_hex(0, 0, 0).lower() == '000000'
+    assert rgb_to_hex(255, 255, 255).lower() == 'ffffff'
+
+
+def test_analogous_color():
+    starting_color = '0000ff'
+    colors = analogous_color(color=starting_color, my_slices=255, slices=255)
+    assert len(colors) == 256
```
ColCarroll: the docstring seems to suggest this should be 255, or whatever value `my_slices` has

mediacloud/mediawords/util/colors.py
```diff
@@ -0,0 +1,116 @@
...
+    existing_colors = set()
+
+    if set_colors is not None:
+        for color in set_colors:
+            existing_colors.add(color)
+
+    # Use the hard coded palette of 25 colors if possible
+    new_color = None
+    for color in __MC_COLORS:
+        if color in existing_colors:
```
ColCarroll: should this say `if color not in existing_colors`?

mediacloud/mediawords/util/colors.py
```diff
@@ -0,0 +1,116 @@
...
+
+    colors = [color]
+    for x in range(1, my_slices + 1):
+        new_color = rotate_color(color_=color, angle_=angle * x)
+        colors.append(new_color)
+
+    return colors
+
+
+def get_consistent_color(db: DatabaseHandler, item_set: str, item_id: str) -> str:
```
ColCarroll: i am nervous about using `item_set` as the name of a string!

mediacloud/mediawords/util/test_colors.py
```diff
@@ -0,0 +1,86 @@
...
+    def test_get_consistent_color(self):
+        partisan_colors = {
+            'partisan_2012_conservative': 'c10032',
+            'partisan_2012_liberal': '00519b',
+            'partisan_2012_libertarian': '009543',
+        }
+
+        for color_id, color in partisan_colors.items():
+            got_color = get_consistent_color(db=self.db(), item_set='partisan_code', item_id=color_id)
+            assert got_color == color
```
ColCarroll: how did this know what color to assign these?

mediacloud/mediawords/util/colors.py
pypt: What parameter name would you suggest? Those two parameters (`item_set` and `item_id`) are freeform strings that will get stored into [`color_sets`](https://github.com/berkmancenter/mediacloud/blob/master/schema/mediawords.sql#L664-L669) table so that next time the function gets called with the same parameters, it will return the same color.

mediacloud/mediawords/util/colors.py
pypt: It should, thank you, good catch!

mediacloud/mediawords/util/test_colors.py
pypt: Sorry, mindlessly copied [`Color::Mix`'s comment](https://metacpan.org/pod/Color::Mix#analogous()). Updated docstring in an attempt to clarify things a little.

mediacloud/mediawords/util/test_colors.py
pypt: It appears that those colors were prefilled in [`mediawords.sql`](https://github.com/berkmancenter/mediacloud/blob/master/schema/mediawords.sql#L673-L676) when originally creating a schema. Added a comment that mentions that.

pypt:
Thanks for the review Colin, merging.

PR #257, opened 2018-01-08
Rewrite language identification code to Python

### LINKED ISSUE #256: Rewrite language identification code to Python
https://github.com/berkmancenter/mediacloud/blob/master/lib/MediaWords/Util/IdentifyLanguage.pm

pypt:
As a (possible) dependency for #215, I've rewritten language identification code to Python.

Fixes #256.

In Perl code, we've been using [`Lingua::Identify::CLD`](https://metacpan.org/pod/Lingua::Identify::CLD), a wrapper around the first version of Chromium [Compact Language Detector](https://github.com/mikemccand/chromium-compact-language-detector) (CLD1). I haven't found any working CLD1 libraries on PyPI (the only one available, [`chromium_compact_language_detector`](https://pypi.python.org/pypi/chromium_compact_language_detector/0.2), is broken on Python 3), so I've made us use CLD version 2 (CLD2) through [`cld2-cffi`](https://pypi.python.org/pypi/cld2-cffi/0.1.4) (the only such module that builds and has unit tests).

Naturally, with CLD2 there are small differences in language detection results visible in 35a57238a5dbe01a16ec9c55ad8d95ad76d44190 and https://github.com/berkmancenter/mediacloud/compare/langid_python?expand=1#diff-d8125f3031a470636490e8f2b70c886a. For the crawler test, about 5% of the sentences get identified as being of a different language than they were with CLD1. CLD2 seems to do slightly worse with very short sentences (e.g. it's not able to identify a sentence *По русскй* -- literally meaning "in the Russian (language)" -- as being Russian), isn't able to recognize Chinese in some cases, and returns different ISO 639 language codes for some (unsupported) languages, e.g. `iw` instead of `he` for Hebrew.

I've looked into a superpopular Python module [`langid`](https://github.com/saffsd/langid.py) too, but I found out that CLD (any version of it) is more accurate. Also, `langid` implements all of its logic in native Python (even the model is part of the [2,41 MB source](https://github.com/saffsd/langid.py/blob/master/langid/langid.py) so it's slower than CLD). Lastly, `langid` doesn't support Hausa (for which we maybe should consider dropping the support altogether, but that's another topic).

Google has also released [CLD3](https://github.com/google/cld3) but there doesn't seem to be any working Python bindings available ([`cld3`](https://pypi.python.org/pypi/cld3/0.2.1) doesn't build and has no tests). We could build our own bindings or try to fork + fix the ones that are available (the developer doesn't seem to be too responsive though).

To test the language identification code, every concrete `AbstractLanguage` implementation now requires `sample_sentence() -> str` method to be implemented which returns a sample sentence for a given language. Then, the unit test iterates over `enabled_languages()` and tests whether the language identification result of a sample sentence for every enabled language matches said language's ISO 639 code (to catch situations such as the one with Hebrew). Not sure if this is the best way to test it (having an abstract method solely used for testing is not quite by the book), but I didn't want to duplicate our list of supported languages in the language identification test, so here we have it.


pypt:
Thank you for the review, merging.

PR #258, opened 2018-01-09
remove one-job-in-queue restriction for public users

### LINKED ISSUE #248: new topics are "created but not queued" for research team
This is happening with alarming regularity to our researchers (Natalie/Fernando/Anushka).  They create a topic using the Topic Mgr UI and then get a red alert at the top saying "topic is created but not queued".  [My code shows this as an error](https://github.com/mitmedialab/MediaCloud-Web-Tools/blob/736dd1ea37cdac67b354dd2370dbb4e0aca219da/src/components/topic/TopicContainer.js#L81).  Is this a normal state in the workflow?

For instance, this just happened to Natalie on topic 1902, which is listed as being in the `mc` queue. To try and fix this she clicked the "spider" button I show there in Topic Mgr, which calls the `topic/spider` API endpoint to try and kick it in to the queue.  But nothing changed. So then she went to the old UI interface and clicked the "Spider" button on the far right.  Then it stared spidering!  :-(

A few questions this provokes:
* why didn't it get queued to start with?
* why is does the "spider" button on the old interface produce different results?

We need to fix this ASAP, because it is making the workflow terrible for our team.  I can't reproduce this myself when I create a topic.

hroberts:
This fixes #248 by removing the one-queued-or-running-job-per-public-user restriction on topic spidering.

The problem is that I had a limit of only one queued topic job for a public user.  The way this was implemented was just to check in topic/spider whether owner of the requested topic is a public user (non admin) and whether that owner has any other topics with pending jobs (in running or queued state).  If that is the case, the end point just exits and returns the existing job state for the other topic.

In theory, the idea was to prevent any one public user from flooding the public topic spider queue with pending jobs.  In practice, this meant that topics would get created but not queued and would not be queueable until some other job got finished running.  The other confusing thing is that this logic was driven by the original owner of the topic, so even admin users could end up having a topic initially created by a non-admin user get stuck in this state.

Since adding the public topic support, I increased the size of the public topic job pool to 4, and it does not seem to have added too much load to mccore1, so I think it's fine to keep that larger public job pool.  That larger job pool also means that the cost of a single user adding several jobs to the pool is not that bad.


pypt:
Wasn't asked for a review, but looks good to me.

PR #271, opened 2018-01-26
Add flake8, correct errors

### LINKED ISSUE #206: Automated Python linting
Marking an issue here to add automated linting to the python code base to keep code conventions consistent.  [Pylint](https://www.pylint.org/) is quite configurable, and easy to get working locally and on travis.  It also has a surprising ability to catch subtle bugs some of the time.

ColCarroll:
Fixes #206 

Adds an automated code linter. I would rather use pylint which does a bit more aggressive checking, but there is an issue related to our symlinking the `snowball` project (see https://github.com/PyCQA/pylint/issues/1470. The fix works,  but is not worth maintaining a fork of `astroid`). I am also ignoring the code generated from `snowball`.

### DIFF
--- a/ansible/roles/python-dependencies/files/requirements.txt
+++ b/ansible/roles/python-dependencies/files/requirements.txt
@@ -19,6 +19,9 @@ CyHunspell==1.2.1
 # Guessing publication dates of stories
 date_guesser==2.1.1
 
+# Code conventions
+flake8==3.5.0
+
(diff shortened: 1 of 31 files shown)


pypt:
Thanks Colin, looks good to me too!

Before merging, could you please add a note about linting in `doc/coding_guidelines.markdown`?

Issue #276, opened 2018-01-29
topics seeded with very large queries time out


hroberts:
topics seeded with queries that take longer than 15 minutes to complete timeout due to the 15 minute solr query timeout.  in the past we have manually run date chunked seed queries to run these topics.

PR #277, opened 2018-01-29
chunk topic spider seed query by month

hroberts:
I'm trying to run an update of the big Trump First Year topic, but the seed query times out after 15 minutes because the solr querying module in perl has a reasonable 15 minute timeout.  For reference, the current trump first year seed query returns 2.2 million stories.  

I could increase that timeout, but that would run into the timeouts set at various other places in the stack, and I don't think it is good practice to rely on > 15 minute solr queries to run successfully over the network.  

To get these queries to run in the past, I have manually edited the query to include only three months, run the spider long enough to import the stories, edited the query for the next three months, run the spider again, and so on.  This is obviously a pain and only allows me to run very large topics.

This patch chunks the topic seed query into months to make it much less likely that a query will be large enough to time out.

I am approving this pull request immediately because I need to get the trump first year topic as soon as possible to make the results available within the book deadline.

### DIFF
--- a/lib/MediaWords/TM/Mine.pm
+++ b/lib/MediaWords/TM/Mine.pm
@@ -32,6 +32,8 @@ use HTML::Entities;
 use List::Util;
 use Parallel::ForkManager;
 use Readonly;
+use Time::Piece;
+use Time::Seconds;
 use URI;
 use URI::Escape;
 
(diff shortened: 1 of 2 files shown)


hroberts:
fixed #276

PR #278, opened 2018-01-30
Generate Python test coverage report and submit it to Coveralls

pypt:
We've been tracking test coverage for our Perl code for quite a while and submitting it to [Coveralls.io](https://coveralls.io/github/berkmancenter/mediacloud), however our Python code's coverage wasn't being reported. Colin suggested that we should track that too.

Coveralls.io doesn't support merging two or more coverage reports for different languages server-side, but I found a way to do it client-side, by generating two "Coveralls.io-compatible JSON" reports and using `coveralls --merge`.

* `./script/run_test_suite.sh` runs both Perl and Python tests. If `MC_TEST_SUITE_REPORT_COVERAGE` environment variable is set, the script generates coverage reports for all of those tests (`.coverage` file for Python tests and `cover_db/` directory for Perl tests).
* After the test run, [`Devel::Cover`](http://search.cpan.org/~pjcj/Devel-Cover-1.29/lib/Devel/Cover.pm)'s `cover` tool generates JSON report of Perl tests and saves it into a file using `Devel::Cover::Report::CoverallsJSON` package (modified version of [`Devel::Cover::Report::Coveralls`](https://metacpan.org/pod/Devel::Cover::Report::Coveralls) which stores the report to a file instead of submitting it to Coveralls.io right away; committed into our repo). Perl's coverage report gets stored in `cover_db/coveralls.json`.
* After the test run, [`coveralls`](https://pypi.python.org/pypi/coveralls) tool generated JSON report of Python tests and saves it into a file. Python's coverage report gets stored in `coverage.json`.
* Lastly, we submit both reports by merging them first using `coveralls --merge`.

Sample build on Coveralls.io: https://coveralls.io/builds/15284361


hroberts:
looks great to me.

script/run_test_suite.sh
```diff
@@ -31,11 +39,28 @@ if [[ "$system_hostname" =~ ^[0-9]+$ || "$shell_hostname" =~ ^[0-9]+$ ]] ; then
...
+if [ ! -z ${MC_TEST_SUITE_REPORT_COVERAGE+x} ]; then
+
+    echo "Removing old test coverage reports..."
+    rm -rf cover_db/ .coverage coverage.json
+
+    # Enable Perl's Devel::Cover
+    export HARNESS_PERL_SWITCHES='-MDevel::Cover=+ignore,t/,+ignore,\.t$'
+
+    # Enable Python's pytest coverage
+    PYTEST_ARGS="--cov=mediacloud/ --cov-report=term --cov-config=.coveragerc"
```
ColCarroll: - should this be `--cov=mediacloud/mediawords` (right now we are picking up the vendored `snowball` library)? 
- i think `--cov-report=term` is the default if you want to save some space
- we can save some file clutter by using `setup.cfg` instead of `.coveragerc` (though if we merge this first, i can make that change once the linting PR lands)

lib/Devel/Cover/Report/CoverallsJSON.pm
```diff
@@ -0,0 +1,175 @@
+package Devel::Cover::Report::CoverallsJSON;
```
ColCarroll: is this the right way to use a perl library?  (i.e., copy/pasting it)

lib/Devel/Cover/Report/CoverallsJSON.pm
```diff
@@ -0,0 +1,175 @@
...
+    {
+        $config = YAML::LoadFile( $CONFIG_FILE );
+    }
+
+    my $json = {};
+
+    # $json->{repo_token} = $config->{repo_token} if $config->{repo_token};
+    # $json->{repo_token} = $ENV{COVERALLS_REPO_TOKEN} if $ENV{COVERALLS_REPO_TOKEN};
+
+    my $is_travis;
```
ColCarroll: if we are using this file, should we simplify it to our use case? for example, get rid of comments, get rid of this branching logic for different services.

.coveragerc
```diff
@@ -0,0 +1,2 @@
+[report]
+omit = */test_*.py
```
ColCarroll: probably add `mediacloud/mediawords/test/*` to this

pypt: In some cases tests are being tested too ;), e.g. we test the `HashServer` class that we use for mock HTTP servers.

lib/Devel/Cover/Report/CoverallsJSON.pm
pypt: Well, I just took the source code from https://metacpan.org/pod/Devel::Cover::Report::Coveralls, modified it a single line, added attribution and put it there. Alternative would have been to release our own CPAN module which I didn't bother to do due to how temporary our use of this code is intended to be.

script/run_test_suite.sh
pypt: Fixed some redundant arguments, thank you. Not sure why we'd ever need `setup.cfg` because Media Cloud code is not a Python module (or could / should be made into one as a whole).

.coveragerc
ColCarroll: cool! just saw it included on coveralls with not much coverage and thought it might be a mistake

script/run_test_suite.sh
ColCarroll: Hrm... I think you're thinking of `setup.py`, which is different, but also a good idea!  `setup.cfg` is an attempt to standardize where configuration goes for various tools to prevent a proliferation of dotfiles.  See, for example, https://github.com/pymc-devs/pymc3/blob/master/setup.cfg

pypt: OK, should we put `setup.cfg` it in `mediacloud/` or in the root directory?

pypt: I've tried to move coverage configuration properties to `setup.cfg` (which is a good idea) but it didn't work. It seems that somehow `pytest` is able to find and `pytest`-related properties from `setup.cfg` but not `coverage` ones - `[coverage:report]`, `[tool:coverage:report]`, `[report]` and other variations that I've tried weren't being read. So, after 30 minutes I just gave up.

ColCarroll:
Frustrating! Looks like it was a bug in a minor release: https://github.com/pytest-dev/pytest-cov/issues/180  

I've got a tiny follow-up that bumps versions and consolidates configuration.

PR #284, opened 2018-02-01
Word2vec model generator classes, worker, and API calls

### LINKED ISSUE #215: generate and store topic-snapshot word2vec models

pypt:
Fixes #215.

**Model generation:**

* [`train_word2vec_model()`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/__init__.py#L16-L64) trains the model using sentences from a specified [sentence iterator](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/sentence_iterators.py). There are iterators for both [topics](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/sentence_iterators.py#L73-L95) and [snapshots](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/sentence_iterators.py#L98-L120). Function stores the generated model into a provided [model store](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/model_stores.py#L12-L31), basically our key-value store; here's such store for [topic models](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/model_stores.py#L102-L117) and [snapshot models](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/word2vec/model_stores.py#L120-L135).
* [`load_word2vec_model`](FIXME) loads the model stored in the model store and returns it as `bytes`.
* Reasonable example of how this all works can be found in [a test](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/mediacloud/mediawords/util/test_word2vec.py#L75-L100)  or [worker code](https://github.com/berkmancenter/mediacloud/tree/word2vec_topic_model/mediacloud/mediawords/job/word2vec).

**Model storage:**

* Models are being Pickled and stored in PostgreSQL's `BYTEA` columns; I figured that they're going to be small enough to fit there every time.
* For every type of model, there are two tables used for storing it: one for metadata (creation date and a reference column to topic / snapshot), the other one for actual raw serialized data for the model (and a reference to the metadata table). This way we can reuse our key-value store code (which insists on object IDs being unique) and decide to store models on S3, cached S3, … at a later date without much of a code change. Schema: [topic](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/schema/mediawords.sql#L2824-L2848), [snapshot](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/schema/mediawords.sql#L1921-L1945).
* Only `KeyedVectors` (`model.wv`) gets stored in the database because we treat those models as immutable, and it conserves space.

**Workers:**

* [Two Python workers](https://github.com/berkmancenter/mediacloud/tree/word2vec_topic_model/mediacloud/mediawords/job/word2vec) do all the model generation.
* There are [stub workers](https://github.com/berkmancenter/mediacloud/tree/word2vec_topic_model/lib/MediaWords/Job/Word2vec) in Perl for us to be able to add model generation jobs from Perl code.
* An appropriate model generation jobs gets added after topic mining job completes, or a snapshot gets created.

**API:**

* Given that Rahul's / Cindy's end will be essentially synchronizing a list of topic / snapshot models, I figured that instead of "models created since date" call we can afford to just list all created models in [`topics/list`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/doc/api_2_0_spec/topics_api_2_0_spec.md#example-6), `topics/single` and [`snapshots/<topics_id>/list`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/doc/api_2_0_spec/topics_api_2_0_spec.md#example-27) so it's easier to keep them in sync (no need to keep track of the "last time we fetched models" date).
* [`topics/<topics_id>/generate_word2vec_model`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/doc/api_2_0_spec/topics_api_2_0_spec.md#topicstopics_idgenerate_word2vec_model-get) and [`snapshots/<snapshots_id>/generate_word2vec_model`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/doc/api_2_0_spec/topics_api_2_0_spec.md#snapshotssnapshots_idgenerate_word2vec_model-get) add the model generation job for a specified topic or snapshot. Not sure if we need those API calls though as those jobs will be added by the topic mining / snapshot creating workers; @rahulbot should I keep them?
* [`topics/<topics_id>/word2vec_model/<models_id>`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/doc/api_2_0_spec/topics_api_2_0_spec.md#topicstopics_idword2vec_modelmodels_id-get) and [`snapshots/<snapshots_id>/word2vec_model/<models_id>`](https://github.com/berkmancenter/mediacloud/blob/word2vec_topic_model/doc/api_2_0_spec/topics_api_2_0_spec.md#snapshotssnapshots_idword2vec_modelmodels_id-get) allow one to download a Pickled model.

**Misc.:**

* Removed topics API spec section on nonexistent `snapshots/single` call.
* Removed all topics-related documentation from main API spec because it's duplicated (and seems to be more up-to-date) in the topics API spec.
* Removed "DONE" annotations in the topics API spec which were apparently used by someone to mark the API calls as "done" while implementing them.
* Removed all sections with "TODO" annotations as I saw no point in keeping them there now that the topics API spec is public.
* Did some more formatting of the topics API spec.


pypt:
Also, @rahulbot you'll loading the models using Python 2, right? Currently models get serialized using [`pickle`'s protocol version 4](https://docs.python.org/3/library/pickle.html#data-stream-format) which is incompatible with Python 2, so I can change it to protocol version 2 if need be.

pypt:
> Yes, I’ll need them compatible with Python 2.7.

Set Pickle protocol to version 2.

> is a topic’s “completes” state blocked by model generation? If not, do I get some status indicating that the model is building?

No, after mining a topic it just adds a separate word2vec job to be completed whenever.

> if model jobs are always generated by snapshots I don’t think we need another endpoint to generate them

OK, removed `/generate_word2vec_model` API call for both topics and snapshots.

pypt:
Models for topics are being generated from story sentences from topic's stories in `topic_stories` (I guess one could say that they represent "live topics"?). Models for (topic) snapshots are being generated from story sentences from snapshot's stories in `snap.stories`.

Do you want to generate models for topics only, or for both topics and their snapshots? I've tried to confirm that with you in https://github.com/berkmancenter/mediacloud/issues/215#issuecomment-356400451, but maybe misunderstood your answer.

*(Will merge when we figure this out.)*

pypt:
Thanks for clarification, removed relevant code. Now models will be generated for snapshots only.

pypt:
Would love to merge and deploy, but Travis keeps on randomly failing, either on `compile.t` with a segfault or on `AbstractJob.t` by not finding a single test in it:

https://travis-ci.org/berkmancenter/mediacloud/jobs/343531044#L3591

Are we running out of RAM here or something?

pypt:
Testing issues taken care of, finally ready for merging, waiting for the conference to be over.

Issue #279, opened 2018-01-30
save audience data into DB


rahulbot:
We need and architecture for storing the audience data from SimilarWeb in the database.  I think we decided that people can only see the freshest number, but we will store the historical data for our internal research purposes. General agreement was that "freshest data" means the monthly average visits over the last 6 months.

pypt:
Could someone give an example of the audience data that is to be stored? Is it a single integer, a list of integers, a free-form JSON or what?

rahulbot:
My naive approach:

- [ ] Add a "current_audience" int field to the media table, which will hold the latest monthly hit total (based on an average over the last 6 months).  This is the number we show publicly.  Not sure if we need a date attached to this or not (so we can say "as of ...").
- [ ] Add an "audience_history" table with "date" (year and month), "size" (int), and "media_id" (foreign key) fields, which have one entry per month for each source that we get from similar web.  This will let us keep historical monthly hit counts around for internal research.

ColCarroll:
- The raw data is below.  List of floats!
- I plan to have a separate table with schema
```
CREATE TABLE similarweb_metrics (                                          
    similarweb_metrics_id  SERIAL                   PRIMARY KEY,           
    domain                 VARCHAR(1024)            NOT NULL,              
    month                  DATE,                                           
    visits                 INTEGER,                                        
    update_date            TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW() 
);                                                                         
```
There is a problem I discussed with Hal about matching to a media id, which is twofold:
– similarweb has data only for top level domains, not subdomains
– (less serious) We cannot look up media ids by url (#214)

It seems sensible to me to stash the raw data, and then decide how to match those URLs to existing media separately.

```
{
  "meta": {
    "request": {
      "granularity": "Monthly",
      "main_domain_only": false,
      "format": null,
      "domain": "realitycheck753.wordpress.com",
      "start_date": "2016-05-01",
      "end_date": "2017-10-31",
      "limit": null,
      "country": "world"
    },
    "status": "Success",
    "last_updated": "2017-10-31"
  },
  "visits": [
    {
      "date": "2016-05-01",
      "visits": 50576.26736803504
    },
    {
      "date": "2016-06-01",
      "visits": 916.6117254286585
    },
    {
      "date": "2016-07-01",
      "visits": 482.3237905072105
    },
    {
      "date": "2016-08-01",
      "visits": 717.8794171247773
    },
    {
[...]

PR #294, opened 2018-02-09
Add similarweb audience metrics

ColCarroll:
Addresses #279

### DIFF
--- a/mediacloud/mediawords/similarweb/similarweb.py
+++ b/mediacloud/mediawords/similarweb/similarweb.py
@@ -1,8 +1,11 @@
 from http import HTTPStatus
-import re
 from urllib.parse import urlparse
 
+import backoff
 import requests
+
+from mediawords.util.url import normalize_url_lossy
+from mediawords.util.web.user_agent import DETERMINED_HTTP_CODES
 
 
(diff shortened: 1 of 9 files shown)


mediacloud/mediawords/similarweb/similarweb.py
```diff
@@ -0,0 +1,197 @@
...
+            url : str
+                URL to make a GET request to
+            params : dict
+                Query parameters to send along as well
+
+        Returns
+        -------
+        requests.Response
+        """
+        return requests.get(url, params=params)
```
hroberts: Can you please use HTTPAdapter to make this retry for up to 10 minutes in the case of network failure?  As is, a network glitch or briefly down api server will just result in some lines in a log file.  It would be great for us to do a better job monitoring our logs, but regardless it would be best just to be resilient here.

We use the following list of codes for determined retries in our user agent code.  I don't see why we shouldn't just make this constant public so that you can use it directly:

https://github.com/berkmancenter/mediacloud/blob/6fd3b144040c14d4d2cfb20290a2646622fdd075/mediacloud/mediawords/util/web/user_agent/__init__.py#L130

PR #298, opened 2018-02-14
Distributed topic spider - extract links

hroberts:
This pull request moves the code that extracts links from stories during the topic spidering process from the perl Topic::Mine module to the python jobs based mediawords.tm.extract_story_links.

This is the first step toward moving all of the topic spidering code from the existing, mostly single process perl module to jobs based python module.  As an intermediate step, the new python jobs are called from the exiting perl module in chunks of 1000, which just waits around for the python jobs to finish, polling the topic_links table to know when they are done.  It should only take an easy edit of the extract_story_links_job.py package to make this new code work with the fully distributed topic spider.

In addition to that top level deliverable, there are several other changes in dependencies of the new python code:

* Util::HTML is now entirely in python, including the needed html_strip() function, which is now BeautifulSoup based (the old perl regex was too fancy for the python regex engine to handle, but the BS module seems to perform fine according to casual benchmarking).

* DBI::Downloads was entirely translated to python so that we can call store_content, fetch_content, and the various extract functions in python.  I have not plugged the perl side into those python implementations yet because I would like to run the python DBI::Downloads code just in the link extractor code for a while before running it for the whole crawler engine.

* I changed the test case class in mediawords.test.test_database to create a template database from mediawords.sql only once during each the life of each process and to use the postgres 'create mediacloud_test template mediacloud_test_template' functionality to more quickly recreate the test database for each unit test.  On my laptop, this reduced the db setup time for each unit test from about 3 seconds to about 0.5 seconds, and the run time for all python tests from about 140 seconds to about 100 seconds.  This change is included in this pull request because this PR includes lots of individual unit tests that rely on having a fresh database.

* There was a get_boingboing_links() function that was no longer working with current boingboing, so I just removed it.

Also:

* This PR also has a first version of a ThrottledUserAgent that will allow per media throttling of the planned topic_fetch_link() job.  That code is not used anywhere yet.

* There are also some miscellaneous small fixes to make mypy happy with existing code in files that I had to touch for this PR.  Mostly, mypy was consistently unhappy when assigning decode_obj_from_bytes_if_needed()  to a str, so I added a decode_str_from_bytes_if_needed() that made it easier to make mypy happy.


hroberts:
going ahead and merging this because we're in a time crunch to speed up the election topics.  I would still love comments on the PR, though.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+        topic_link = {
+            'topics_id': topic['topics_id'],
+            'stories_id': story['stories_id'],
+            'url': link
+        }
+
+        db.create('topic_links', topic_link)
+
+    db.query(
+        "update topic_stories set link_mined = 't' where stories_id = %(a)s and topics_id = %(b)s",
```
ColCarroll: change `a` to `stories_id` and `b` to `topics_id`.

mediacloud/mediawords/util/html.py
```diff
@@ -1,3 +1,6 @@
+"""Various utility functions for handling html."""
+
+from bs4 import BeautifulSoup
```
ColCarroll: super nit: `pip` installed packages go after system packages, and before local packages (so `BeautifulSoup` goes in its own group after `typing`)

mediacloud/mediawords/util/html.py
```diff
@@ -48,31 +54,32 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
 
         match = re.search(r'content\s*?=\s*?"\d*?\s*?;?\s*?URL\s*?=\s*?\'(.+?)\'', inner_tag, re.I)
         if match:
-            inner_url = match.group(1)
+            inner_url = str(match.group(1))
```
ColCarroll: is this `bytes` otherwise?

mediacloud/mediawords/util/html.py
```diff
@@ -48,31 +54,32 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
...
+                inner_url = str(match.group(1))
             else:
                 # Fallback
                 match = re.search(r'content\s*?=\s*?["\']\d*?\s*?;?\s*?URL\s*?=\s*?(.+?)["\']', inner_tag, re.I)
                 if match:
-                    inner_url = match.group(1)
+                    inner_url = str(match.group(1))
 
-        if is_http_url(inner_url):
+        if is_http_url(str(inner_url)):
```
ColCarroll: it looks like `inner_url` is already cast to string, right?

mediacloud/mediawords/util/html.py
```diff
@@ -34,12 +39,13 @@ def link_canonical_url_from_html(html: str, base_url: str = None) -> Optional[st
...
-def meta_refresh_url_from_html(html: str, base_url: str = None) -> Optional[str]:
+def meta_refresh_url_from_html(html: str, base_url: Optional[str] = None) -> Optional[str]:
     """From the provided HTML, determine the <meta http-equiv="refresh" /> URL (if any)."""
+    def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url: Optional[str]=None) -> Optional[str]:
+        """Given a <meta ...> tag, return the url from the content="url=XXX" attribute.
 
-    def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Optional[str]:
-        """Given a <meta ...> tag, return the url from the content="url=XXX" attribute.  return undef if no such url is
-        found."""
+        return undef if no such url isfound.
```
ColCarroll: return `None` if no such url is found`

mediacloud/mediawords/util/html.py
```diff
@@ -48,31 +54,32 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
...
-                    inner_url = match.group(1)
+                    inner_url = str(match.group(1))
 
-        if is_http_url(inner_url):
+        if is_http_url(str(inner_url)):
             return inner_url
 
         if inner_base_url is not None:
-            return urljoin(base=inner_base_url, url=inner_url)
+            return urljoin(base=str(inner_base_url), url=str(inner_url))
```
ColCarroll: I think everything is already cast here?

mediacloud/mediawords/util/html.py
```diff
@@ -48,31 +54,32 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
...
 
-    html = decode_object_from_bytes_if_needed(html)
-    base_url = decode_object_from_bytes_if_needed(base_url)
+    html = str(decode_object_from_bytes_if_needed(html))
+    base_url_decode = decode_object_from_bytes_if_needed(base_url)
+    base_url = None if base_url_decode is None else str(base_url_decode)
 
-    if not is_http_url(base_url):
-        log.error("Base URL is not HTTP(s): %s" % base_url)
+    if not is_http_url(str(base_url)):
```
ColCarroll: unnecessary `str`

mediacloud/mediawords/util/html.py
```diff
@@ -81,3 +88,107 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
             return url
 
     return None
+
+
+def _sententize_block_level_tags(s: str) -> str:
```
ColCarroll: "sententize"?  😄

mediacloud/mediawords/util/web/user_agent/test_throttled.py
```diff
@@ -0,0 +1,62 @@
...
+from mediawords.test.http.hash_server import HashServer
+from mediawords.util.web.user_agent.throttled import ThrottledUserAgent
+from mediawords.util.web.user_agent.throttled import McThrottledUserAgentTimeoutException
+import mediawords.util.web.user_agent.throttled
+
+
+class TestThrottledUserAgent(TestDatabaseWithSchemaTestCase):
+    """test case for ThrottledUserAgent."""
+
+    def test_request(self) -> None:
```
ColCarroll: it looks like this should be a number of smaller unittests!

mediacloud/mediawords/util/web/user_agent/test_throttled.py
```diff
@@ -0,0 +1,62 @@
...
+        """Test requests with throttling."""
+        pages = {'/test': 'Hello!', }
+        port = 8888
```
ColCarroll: this won't work for anyone (like me!) running a Jupyter server on the default port.  You might look at https://github.com/getsentry/responses or the method used elsewhere for getting a random unused port.

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
...
+from mediawords.util.log import create_logger
+log = create_logger(__name__)
+
+# default amount of time in between requests
+_DEFAULT_DOMAIN_TIMEOUT = 10
+
+
+class McThrottledUserAgentTimeoutException(Exception):
+    """Exception raised when a ThrottledUserAgent request fails to get a domain request lock."""
+
```
ColCarroll: extra newline

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
...
+import mediawords.util.config
+from mediawords.util.web.user_agent import UserAgent
+from mediawords.util.web.user_agent.request.request import Request
+from mediawords.util.web.user_agent.response.response import Response
+
+from mediawords.util.log import create_logger
+log = create_logger(__name__)
+
+# default amount of time in between requests
+_DEFAULT_DOMAIN_TIMEOUT = 10
```
ColCarroll: should this be in a global config?

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
+"""Implement ThrottledUserAgent as a sub class of mediawords.util.web.UserAgent with per domain throttling."""
+
+import typing
+
+import mediawords.db
+import mediawords.util.config
+from mediawords.util.web.user_agent import UserAgent
+from mediawords.util.web.user_agent.request.request import Request
+from mediawords.util.web.user_agent.response.response import Response
+
```
ColCarroll: extra newline

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
...
+        self.domain_timeout = domain_timeout
+
+        if self.domain_timeout is None:
+            config = mediawords.util.config.get_config()
+            if 'throttled_user_agent_domain_timeout' in config['mediawords']:
+                self.domain_timeout = int(config['mediawords']['throttled_user_agent_domain_timeout'])
+            if self.domain_timeout is None or self.domain_timeout < 1:
+                self.domain_timeout = _DEFAULT_DOMAIN_TIMEOUT
+
+        UserAgent.__init__(self)
```
ColCarroll: should probably be `super().__init__()` (unless you're sure this is what you want - let me know if you want elaboration)

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
...
+class McThrottledUserAgentTimeoutException(Exception):
+    """Exception raised when a ThrottledUserAgent request fails to get a domain request lock."""
+
+    pass
+
+
+class ThrottledUserAgent(UserAgent):
+    """Add per domain throttling to mediawords.util.web.UserAgent."""
+
+    def __init__(self, db: mediawords.db.DatabaseHandler, domain_timeout: typing.Optional[int]=None) -> None:
```
ColCarroll: maybe `domain_timeout: int=_DEFAULT_DOMAIN_TIMEOUT`?

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
...
+        If not present in mediawords.yml, use _DEFAULT_DOMAIN_TIMEOUT.
+        """
+        self.db = db
+        self.domain_timeout = domain_timeout
+
+        if self.domain_timeout is None:
+            config = mediawords.util.config.get_config()
+            if 'throttled_user_agent_domain_timeout' in config['mediawords']:
+                self.domain_timeout = int(config['mediawords']['throttled_user_agent_domain_timeout'])
+            if self.domain_timeout is None or self.domain_timeout < 1:
```
ColCarroll: overriding `domain_timeout` for values < 1 isn't documented - I would not silently override parameters.

mediacloud/mediawords/util/web/user_agent/throttled.py
```diff
@@ -0,0 +1,67 @@
...
+        # this function does not use a table lock, so some extra requests might sneak through, but that's better than
+        # dealing with a lock.  we use a postgres function to make the the race condition as rare as possible.
+        got_domain_lock = self.db.query(
+            "select get_domain_web_requests_lock(%s, %s)",
+            (domain, self.domain_timeout)).flat()[0]
+
+        if not got_domain_lock:
+            raise McThrottledUserAgentTimeoutException("domain " + str(domain) + " is locked.")
+
+        return super(ThrottledUserAgent, self).request(request)
```
ColCarroll: python3 allows/encourages `super().request(request)`

tools/benchmark/benchmark_html_strip.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+"""Benchmark mediawords.util.html.html_strip()."""
```
ColCarroll: is this intended to be checked in?  benchmark suites are great – we used https://github.com/airspeed-velocity/asv for a project, and it has been quite useful (see the [results here](http://pandas.pydata.org/speed/pymc3/))

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+    """Parse html content out of javascript.
+
+    Forbes is putting all of its content into a javascript variable, causing our extractor to fall down.
+    this function returns the html assigned to the javascript variable.
+
+    Returns:
+    the original content or the content parsed from the javascript
+
+    """
+    match = re.search(r'.*fbs_settings.content[^\}]*body\"\:\"([^"\\]*(\\.[^"\\]*)*)\".*', content, flags=re.M | re.S)
```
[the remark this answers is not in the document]
hroberts: I agree.  I'm having to balance not going down every rabbit hole with making shinier more robust code as I port.  I have added a separate issue to figure out whether this bit of magic even works any more.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+log = create_logger(__name__)
+
+# PostgreSQL table name for storing raw downloads
+RAW_DOWNLOADS_POSTGRESQL_KVS_TABLE_NAME = 'raw_downloads'
+
+# Min. content length to extract (assuming that it has some HTML in it)
+MIN_CONTENT_LENGTH_TO_EXTRACT = 4096
+
+# these are initialized by calling the various get_*_story() functions below
+_inline_store = None
```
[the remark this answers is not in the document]
hroberts: I poked around in the world of pythonicism for a bit and couldn't find this addressed anywhere.  What is the argument for putting globals instead classes instead of just within packages?  Why mediawords.dbi.downloads.Globals.MIN_CONTENT_LENGTH_TO_EXTRACT rather than mediawords.dbi.downloads.MIN_CONTENT_LENGTH_TO_EXTRACT?

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+from mediawords.db import DatabaseHandler
+from mediawords.key_value_store import KeyValueStore
+import mediawords.util.extract_text
+import mediawords.util.html
+from mediawords.util.log import create_logger
+
+log = create_logger(__name__)
+
+# PostgreSQL table name for storing raw downloads
+RAW_DOWNLOADS_POSTGRESQL_KVS_TABLE_NAME = 'raw_downloads'
```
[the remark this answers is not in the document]
hroberts: i personally find it a lot more confusing to have to go to a different file to find some random constant rather than looking for it at the top of the package I'm writing.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+# these are initialized by calling the various get_*_story() functions below
+_inline_store = None
+_amazon_s3_store = None
+_postgresql_store = None
+_store_for_writing = None
+
+
+class McDBIDownloadsException(Exception):
+    """Default exceptions for this package."""
+
```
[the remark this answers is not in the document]
hroberts: My flake8 gets angry at me if I put anything other than exactly two lines before each top level class or def line.  I just manually verified this.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+
+def _get_inline_store() -> KeyValueStore:
+    """Get lazy initialized database inline store."""
+    global _inline_store
+
+    if _inline_store is not None:
+        return _inline_store
+
+    from mediawords.key_value_store.database_inline import DatabaseInlineStore
```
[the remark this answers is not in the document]
hroberts: You're right.  I'm not sure why we did this even in the perl.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+        raise McDBIDownloadsException("Amazon S3 download store is not configured.")
+
+    store_params = {
+        'access_key_id': config['amazon_s3']['downloads']['access_key_id'],
+        'secret_access_key': config['amazon_s3']['downloads']['secret_access_key'],
+        'bucket_name': config['amazon_s3']['downloads']['bucket_name'],
+        'directory_name': config['amazon_s3']['downloads']['directory_name'],
+    }
+
+    if config['mediawords'].get('cache_s3_downloads', False):
```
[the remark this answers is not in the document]
hroberts: I think for non-critical boolean configuration variables like this it is fine to just accept a default value if there is no configuration value.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+        store_params['cache_table'] = 'cache.s3_raw_downloads_cache'
```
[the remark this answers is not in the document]
hroberts: fixed

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+
+    return r
+
+
+def _set_cached_extractor_results(db, download, results) -> None:
+    """Store results in extractor cache and manage size of cache."""
+    max_cache_entries = 1000 * 1000
+
+    # occasionally delete too old entries in the cache
+    if random.random() * (max_cache_entries / 10) < 1:
```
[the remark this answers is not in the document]
hroberts: I have added more comments to the code to explain what's going on here.  We only need on the order of ten thousand rows in this table to handle its job as a sort of working memory of extractions within the topic spider, but it is pretty trivial for us to host a million-ish row table on our postgres install.  It is expensive (in query time or denormalization  complexity) to directly keep track of a postgres table size, though, so it is simplest and quickest just to trim the table every 100k requests. 

We could use some sort of cache module, but we need this to be in some common data store shared between servers, and postgres is our general purpose robust solution for a common data store.  I could use some sort of caching module that hides this logic, but this is simple, robust, and transparent in a critical code path.

I disagree that it is unintuitive that a set_cache_value() function will be responsible for trimming old data from the cache.  The only alternatives are either requiring the user to periodically run a 'trim_cache()' function or to have a daemon running somewhere to manage the cache, both of which are more complex and fragile.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+        db.query(
+            """
+            delete from cached_extractor_results
+                where cached_extractor_results_id in (
+                    select cached_extractor_results_id from cached_extractor_results
+                        order by cached_extractor_results_id desc offset max_cache_entries )
```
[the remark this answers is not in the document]
hroberts: fixed.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+def _call_extractor_on_html(content: str) -> dict:
+    """Call extractor on the content."""
+    extracted_html = mediawords.util.extract_text.extract_article_from_html(content)
+    extracted_text = mediawords.util.html.html_strip(extracted_html)
+
+    return {'extracted_html': extracted_html, 'extracted_text': extracted_text}
+
+
+def extract_content(content: str) -> dict:
+    """Accept a content_ref pointing to an HTML string.  Run the extractor on the HTMl and return the extracted text.
```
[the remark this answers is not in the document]
hroberts: fixed and added brief definition of extraction:

Extraction means pulling the substantive text out of a web page, eliminating the navigation, ads, and other boilerplate content.

In practice, it currently basically means calling python-readability on the html.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+    """
+    # Don't run through expensive extractor if the content is short and has no html
+    if len(content) < MIN_CONTENT_LENGTH_TO_EXTRACT and re.search(r'<.*>', content) is None:
+        log.info("Content length is less than MIN_CONTENT_LENGTH_TO_EXTRACT and has no HTML so skipping extraction")
+        ret = {'extracted_html': content, 'extracted_text': content}
+    else:
+        ret = _call_extractor_on_html(content)
+
+        # if we didn't get much text, try looking for content stored in the javascript
+        if len(ret['extracted_text']) < 256:
```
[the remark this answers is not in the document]
hroberts: fixed.

mediacloud/mediawords/test/db/__init__.py
```diff
@@ -221,8 +221,9 @@ def create_test_story_stack_numerated(db: DatabaseHandler,
                                       num_stories_per_feed: int,
                                       label: str = 'test'):
     """Call create_test_story_stack with num_media, num_feeds_per_medium, num_stories_per_feed instead of dict."""
-
-    label = decode_object_from_bytes_if_needed(label)
+    if label is None:  # perl can bass undef / None explicitly so that label ends up without default
```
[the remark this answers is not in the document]
hroberts: I'm trying to decide whether I meant the fish or the musical clef here.

mediacloud/mediawords/test/test_database.py
```diff
@@ -12,12 +13,19 @@
...
+
 class TestDatabaseTestCase(TestCase):
-    """TestCase that connects to the test database which is later accessible as self.db()"""
+    """TestCase that connects to the test database which is later accessible as self.db()."""
+
     __db = None
 
     @staticmethod
-    def _create_database_handler():
+    def _create_database_handler() -> None:
```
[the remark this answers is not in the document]
hroberts: fixed.

mediacloud/mediawords/test/test_database.py
```diff
@@ -49,16 +57,54 @@ def db(self) -> DatabaseHandler:
 
 
 class TestDatabaseWithSchemaTestCase(TestCase):
-    """TestCase that connects to the test database and imports schema; database is later accessible as self.db()"""
+    """TestCase that connects to the test database and imports schema; database is later accessible as self.db()."""
 
-    def setUp(self):
+    _template_db_created = False
+
+    def setUp(self) -> None:
```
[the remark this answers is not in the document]
hroberts: yes, that's much cleaner!  fixed.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+    for tag in soup.find_all(href=True):
+        url = tag['href']
+
+        if re.search(_IGNORE_LINK_PATTERN, url, flags=re.I) is not None:
+            continue
+
+        if not mediawords.util.url.is_http_url(url):
+            continue
+
+        url = re.sub(r'www[a-z0-9]+.nytimes', 'www.nytimes', url, flags=re.I)
```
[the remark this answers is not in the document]
hroberts: Fixed the pattern to be more consistent.

We do a lot of this sort of thing, so I just ran a simple benchmark of 'string_a in string_b' vs. re.sub() to get a sense of what the performance difference is.  For matching against link length strings, the differences is small (7.0 vs. 5.7).  For matching against 10k or 100k strings, the difference is much bigger (3.8 vs 1.9).

I'm leaving this as is, because the small performance here is going to be swamped by the cost of running the much bigger regexes above.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if 'youtube' not in url:
+            continue
+
+        if not url.lower().startswith('http'):
+            url = 'http:' + url
+
+        url = url.strip()
+
+        url = re.sub('youtube-embed', 'youtube', url)
```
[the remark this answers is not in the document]
hroberts: agreed.  fixed.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+def get_extracted_html(db: DatabaseHandler, story: dict) -> str:
+    """Get the extracted html for the story.
+
+    We don't store the extracted html of a story, so we have to get the first download assoicated with the story
+    and run the extractor on it.
+
+    """
+    download = db.query(
```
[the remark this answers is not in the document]
hroberts: db.query returns a response object, which has a number of possible methods you can run to get the actual results.  We most commonly use .hashes(), which returns a list of dicts (we should rename that to dicts() for sure!).  But another option is .hash(), which just returns the first dict for common cases like above of wanting to get just the one thing for the database.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+        "select * from downloads where stories_id = %(a)s order by downloads_id limit 1",
```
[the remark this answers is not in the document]
hroberts: I really dislike this:

```
db.query("select * from stories where stories_id = %{stories_id}s", {'stories_id': stories_id})
```

The four redundant 'stories_id' names are hard for my brain to parse.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    extracted_html = get_extracted_html(db, story)
+
+    html_links = get_links_from_html(extracted_html)
+    text_links = get_links_from_story_text(db, story)
+    youtube_links = get_youtube_embed_links(db, story)
+
+    links = html_links + text_links + youtube_links
+
+    links = list(filter(lambda x: re.search(_IGNORE_LINK_PATTERN, x, flags=re.I) is None, links))
```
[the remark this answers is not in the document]
hroberts: fixed.

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+
+    link_lookup = {}
+    for url in links:
+        link_lookup[mediawords.util.url.normalize_url_lossy(url)] = url
+
+    links = link_lookup.values()
```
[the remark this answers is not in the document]
hroberts: fixed

mediacloud/mediawords/util/html.py
hroberts: fixed.

mediacloud/mediawords/util/html.py
hroberts: fixed by removing the str().

I was trying to avoid a mypy typing complaint here, but I was misunderstanding the root of the problem.

mediacloud/mediawords/util/html.py
hroberts: Surprised this is the only of those that snuck through!

mediacloud/mediawords/util/html.py
hroberts: fixed all of these.

mediacloud/mediawords/util/html.py
hroberts: probably bad humor, but trying to get across the idea that I'm making the text more easily sentence parseable.  this used to be called something like 'add_new_lines_around_block_level_tags()', but that doesn't really express the purpose or even what it does.

mediacloud/mediawords/util/web/user_agent/test_throttled.py
hroberts: I don't really buy into the orthodoxy of having lots of little functions testing the smallest possible things.

The big chunk of this is a flow of tests based on the timing of making a single request, doing some tests within the timeout period, waiting for the timeout period to end, and then doing some more tests.  I could write a test case to repeat that timing for each individual test, but that just seems more confusing (and slower!) to me.

mediacloud/mediawords/util/web/user_agent/throttled.py
hroberts: I find it clearer just to have all of the constants at the top of the file unless there's a reason to isloate them.

mediacloud/mediawords/util/web/user_agent/throttled.py
hroberts: fixed.

mediacloud/mediawords/util/web/user_agent/throttled.py
hroberts: I want the first fallback to be to go to mediawords.yml, and only after that to go to DEFAULT_DOMAIN_TIMEOUT.  I can't do that if I put the constant default in the parameter list.

mediacloud/mediawords/util/web/user_agent/throttled.py
hroberts: fixed.

mediacloud/mediawords/util/web/user_agent/throttled.py
hroberts: fixed.  this file is a little mini history of me figuring out python oo syntax!

tools/benchmark/benchmark_html_strip.py
hroberts: looks helpful. will check out.  thanks!

hroberts:
Thanks for all of the comments.  This was very helpful as a python newbie.

I think I have addressed everything.  I've been making these changes in the main distributed_spider_links branch, where I am now finishing up another big chunk of work to add jobs based link fetching / story generation.

hroberts:
Thanks for all of the comments.  This was very helpful as a python newbie.

I think I have addressed everything.  I've been making these changes in the main distributed_spider_links branch, where I am now finishing up another big chunk of work to add jobs based link fetching / story generation.

pypt:
> There are also some miscellaneous small fixes to make mypy happy with existing code in files that I had to touch for this PR. Mostly, mypy was consistently unhappy when assigning decode_obj_from_bytes_if_needed() to a str, so I added a decode_str_from_bytes_if_needed() that made it easier to make mypy happy.

What was mypy's complaint about `decode_obj_from_bytes_if_needed()`?

mediacloud/mediawords/dbi/downloads.py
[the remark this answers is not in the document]
pypt: We could (should) definitely self-contain global variables, but IMHO Hal did the right thing by not doing it now. As per our Perl -> Python rewrite strategy, we rewrite first and refactor later because those are two different tasks, and the latter task is much more easily done while in Python due to various fancy tooling being available to us (for comparison, Perl doesn't even have a proper IDE).

mediacloud/mediawords/dbi/downloads.py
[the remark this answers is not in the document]
pypt: It's not a global constant, it's a local one, intended to be used only by the module. Possible improvement would be to prefix its name with `__` to communicate that fact.

mediacloud/mediawords/dbi/downloads.py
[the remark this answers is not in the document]
pypt: It has a default value (`False`) set by our default configuration `mediawords.yml.dist` so it's safe to just access `config['mediawords']['cache_s3_downloads']` here.

mediacloud/mediawords/dbi/downloads.py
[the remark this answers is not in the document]
pypt: We already have a PostgreSQL-based object cache that's being purged by a separate `purge_object_caches.py` script run by Supervisor, so it would be nice to reuse that cache here somehow. But again, to be able to do that, we want this and other code to be in Python first!

pypt:
> DBI::Downloads was entirely translated to python so that we can call store_content, fetch_content, and the various extract functions in python. I have not plugged the perl side into those python implementations yet because I would like to run the python DBI::Downloads code just in the link extractor code for a while before running it for the whole crawler engine.

Makes me slightly nervous to have duplicate code in two implementations which might diverge from each other soon. Are you testing this rewritten module that way? What are the prerequisites for completely replacing Perl implementation of `::DBI::Downloads` to the Python implementation? Can I do this?

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+    '(www.addtoany.com)|(novostimira.com)|(ads\.pheedo)|(www.dailykos.com\/user)|'
+    '(livejournal.com\/(tag|profile))|(sfbayview.com\/tag)|(absoluteastronomy.com)|'
+    '(\/share.*http)|(digg.com\/submit)|(facebook.com.*mediacontentsharebutton)|'
+    '(feeds.wordpress.com\/.*\/go)|(sharetodiaspora.github.io\/)|(iconosquare.com)|'
+    '(unz.com)|(answers.com)|(downwithtyranny.com\/search)|(scoop\.?it)|(sco\.lt)|'
+    '(pronk.*\.wordpress\.com\/(tag|category))|(wn\.com)|(pinterest\.com\/pin\/create)|(feedblitz\.com)|(atomz.com)')
+
+
+def get_links_from_html(html: str) -> typing.List[str]:
+    """Return a list of all links that appear in the html.
```
pypt: `html = decode_object_from_bytes_if_needed(html)`

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+            continue
+
+        url = re.sub(r'www[a-z0-9]+.nytimes', 'www.nytimes', url, flags=re.I)
+
+        links.append(url)
+
+    return links
+
+
+def get_youtube_embed_links(db: DatabaseHandler, story: dict) -> typing.List[str]:
```
pypt: `story = decode_object_from_bytes_if_needed(story)`

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+
+        url = re.sub('youtube-embed', 'youtube', url)
+
+        links.append(url)
+
+    return links
+
+
+def get_extracted_html(db: DatabaseHandler, story: dict) -> str:
+    """Get the extracted html for the story.
```
pypt: `story = decode_object_from_bytes_if_needed(story)`

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+    download = db.query(
+        "select * from downloads where stories_id = %(a)s order by downloads_id limit 1",
+        {'a': story['stories_id']}).hash()
+
+    extractor_results = mediawords.dbi.downloads.extract(db, download, use_cache=True)
+    return extractor_results['extracted_html']
+
+
+def get_links_from_story_text(db: DatabaseHandler, story: dict) -> typing.List[str]:
+    """Get all urls that appear in the text or description of the story using a simple regex."""
```
pypt: `story = decode_object_from_bytes_if_needed(story)`

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+    links = []
+    for url in re.findall(r'https?://[^\s\")]+', story_text):
+        url = re.sub(r'\W+$', '', url)
+        links.append(url)
+
+    return links
+
+
+def get_links_from_story(db: DatabaseHandler, story: dict) -> typing.List[str]:
+    """Extract and return linksk from the story.
```
pypt: `story = decode_object_from_bytes_if_needed(story)`

mediacloud/mediawords/tm/extract_story_links.py
```diff
@@ -0,0 +1,196 @@
...
+    for url in links:
+        link_lookup[mediawords.util.url.normalize_url_lossy(url)] = url
+
+    links = link_lookup.values()
+
+    return links
+
+
+def extract_links_for_topic_story(db: DatabaseHandler, story: dict, topic: dict) -> None:
+    """
```
pypt:
```python
story = decode_object_from_bytes_if_needed(story)
topic = decode_object_from_bytes_if_needed(topic)
```

mediacloud/mediawords/util/html.py
```diff
@@ -81,3 +88,107 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
...
+def html_strip(s: str, include_title: bool=False) -> str:
+    """Strip the html tags, html comments, any any text within TITLE, SCRIPT, APPLET, OBJECT, and STYLE tags.
+
+    Derived from code by powerman from: http://www.perlmonks.org/?node_id=161281.
+
+    Arguments:
+    s - html to strip
+    include_title - if true, the title text in the returned text
+    """
+    s = str(decode_object_from_bytes_if_needed(s))
```
pypt: 1) No need for `str()`
2) Please see previous note about `bool`s

mediacloud/mediawords/util/html.py
```diff
@@ -81,3 +88,107 @@ def __get_meta_refresh_url_from_tag(inner_tag: str, inner_base_url=None) -> Opti
...
+    for tag in soup(remove_tags):
+        tag.decompose()
+
+    text = soup.get_text(' ', strip=True)
+
+    return text.strip()
+
+
+def html_title(html: str, fallback: str, trim_to_length: int=0) -> Optional[str]:
+    """Parse the content for tags that might indicate the story's title.
```
pypt: 1)

```python
html = decode_object_from_bytes_if_needed(html)
fallback = decode_object_from_bytes_if_needed(fallback)
```

2) Maybe add an extra `int()` around `trim_to_length` for a good measure.

mediacloud/mediawords/util/web/user_agent/test_throttled.py
pypt: Added where? Can't find it.

Why not just use `random_unused_port()` from `mediawords.util.network`?

t/data/crawler_stories/gv/9.json
```diff
@@ -1,548 +1,537 @@
...
...\t<span class=\"comment-date\">2009-06-8 at 17:27 pm</span>\t\t\t\t</div>\n\t\t\t\t<div class=\"reply\">\n\t\t\t\t\t<a rel='nofollow' class='comment-reply-link' href='/2009/06/08/hong-kong-say-no-to-cctvb/?replytocom=1569791#respond' onclick='return addComment.moveForm(\"comment-1569791\", \"1569791\", \"respond\", \"78992\")'>Reply to this</a>\t\t\t\t</div>\n\t\t\t</div>\n\t\n<ul class='children'>\n\t\n\t\t<li class=\"comment byuser comment-author-oiwan bypostauthor odd alt depth-2\" id=\"li-comment-1570040\">\n\t\t\t<div class=\"commentdiv\" id=\"comment-1570040\">\n\t\t\t\t<div class=\"comment-author vcard\">\n\t\t\t\t\t<img alt='' src='http://www.gravatar.com/avatar/6f109828da91ea546c9b68eb56fed073?s=48&amp;d=http%3A%2F%2Fwww.gravatar.com%2Favatar%2Fad516503a11cd5ca435acc9bb6523536%3Fs%3D48&amp;r=G' class='author-profile-pic avatar avatar-48 photo' height='48' width='48' />\t\n\t\t\t\t\t<cite class=\"comment-author\">\n\t\t\t\t\t\t<a href='/author/oiwan/' class='url'>Oiwan Lam</a>\t\t\t\t\t</cite>\n\t\t\t\t</div>\n\t\t\t\t\t\n\t\t\t\t<p>yeah, let&#8217;s celebrate corruption instead.</p>\n\t\n\t\t\t\t<div class=\"comment-meta commentmetadata\">\n\t\t\t\t\t<a href=\"http://globa
```
pypt: New `html_strip()` seems to mess up periods (`:` becomes `.`, `...` becomes `..`), add a period after every newline, add some extra spaces around periods. Paragraphs are gone too, everything seems to be put into a single line. Not sure what impact will it have for sentence tokenization.

mediacloud/mediawords/dbi/downloads.py
```diff
@@ -0,0 +1,388 @@
...
+    # parsing the feed afterward.  so we want to keep the feed_error state even if we redownload
+    # the content
+    new_state = 'success' if download['state'] != 'feed_error' else 'feed_error'
+
+    try:
+        path = _get_store_for_writing().store_content(db, download['downloads_id'], content)
+        error = ''
+    except Exception as e:
+        raise McDBIDownloadsException("error while trying to store download %d: %s" % (download['downloads_id'], e))
+        new_state = 'error'
```
pypt: screenshot 2018-02-20 05 23 28

Row in `downloads` will never get its `state` set to `error`.

PR #305, opened 2018-02-19
Partition "stories_tags_map" table

pypt:
**Problem:**

Primary key for `stories_tags_map` exceeded `INT`'s max. value; also, reindexing and readding foreign keys takes a lot of time.

**Solution:**

* Make primary key a `BIGINT`.
* Partition table to accommodate for further growth and make it easier to readd foreign keys.

Partitioning scheme closely follows what we have for `bitly_total_clicks`:

* `stories_tags_map` master table without no primary keys, foreign keys or indexes.
    * As the master table has no indexes, `ANALYZE SELECT ...` will print its intention to do a sequential scan over the table. No need to fear, the master table has no rows to scan.
* `stories_tags_map_partition_chunk_size()` function - returns 100m, this is how many `stories_id` chunks the table will be split into, i.e. stories with `0 <= stories_id < 100,000,000` will go into partition `stories_tags_map_00`, stories with `100,000,000 <= stories_id < 200,000,000` into partition `stories_tags_map_01`, ...
* `stories_tags_map_create_partitions()` - creates all the needed partitions for `MAX(stories_id) FROM stories` if they don't exist already; called in `mediawords.sql` and get called periodically in `create_missing_partitions()` (which, in turn, gets called by the `tools/db/create_missing_partitions.py` script).
* `stories_tags_map_partition_by_stories_id_insert_trigger()` trigger function - upserts entries into `stories_tags_map`, i.e. doesn't throw an exception if the tag mapping that is being INSERTed already exists.

**Misc.:**

* `table_exists()` now supports testing tables with schema prepended, e.g. `public.stories`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,10 +56,7 @@ env:
         # Superglue test feed URL:
         #
         # * MC_SUPERGLUE_TEST_URL
-        #
-        # (Disabled because Superglue live feeds are currently down)
-        #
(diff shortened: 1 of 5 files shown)


pypt:
Foreign keys to both `stories` and `tags` added and validated.

PR #318, opened 2018-02-24
Python handler deproxy

hroberts:
Substantive comments copied from first commit below.  I'm going to go ahead and merge this once it passes the test suite because it is causing topic blocking bugs in production.  But I'm eager for a review for how this impacts the database handling and proxying.

Don't deep_copy or wantarray return value in HandlerProxy if the caller is python

When a database handle was created in perl and passed to python, the hashes() and flat() methods were returning the single value rather than a list in python when the query returned only one row.  This is different than the behavior when calling those functions from a python created db handle, so it was causing intermittent havoc in python functions called from perl with a perl db handle.  The cause was that those methods were calling wantarray in perl to figure out whether to return a list or a ref to a list, which was causing the return type confusion.

You can reproduce this issue by uncommenting the new code in HandlerProxy hashes() and running lib/MediaWords/DB/t/HandlerProxy.t

In the theory of reducing unnecessary work and round trip complexity, I added a check to all of the deep_copy( $return_value ) calls in HandlerProxy.pm to avoid the deep_copy if the caller is python.

### DIFF
--- a/lib/MediaWords/DB/HandlerProxy.pm
+++ b/lib/MediaWords/DB/HandlerProxy.pm
@@ -112,8 +112,10 @@ sub schema_is_up_to_date($)
 
     sub columns($)
     {
-        my $self           = shift;
-        my $return_value   = $self->{ _python_result }->columns();
+        my $self         = shift;
+        my $return_value = $self->{ _python_result }->columns();
+        return $return_value if MediaWords::Util::Python::called_from_python();
+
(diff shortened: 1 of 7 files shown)


pypt:
`wantarray` is there because some of our code might still expecting array instead of arrayref from database results, so I've ported that from `DBIx::Simple` directly. I think I got rid of `@array = $db->something` cases, but not sure.

Good change IMHO.

PR #322, opened 2018-03-05
Remove Bit.ly support

### LINKED ISSUE #274: Drop Bit.ly support?
@hroberts, should we drop Bit.ly support, given that the click counts aren't too useful in active research?

Benefits:

* One less thing to think about
* We'd get rid of a chunk of untested code (_as it so happens, I managed to cover most of the API client with tests in `bitly_python` branch recently, but it would make me even happier to not have to rewrite it to Python, so I have no "sentiments" of that (or any other of my) code)_
* We wouldn't be really deleting anything: removed code could be restored from the Git history, and raw JSON objects are easy to re-upload to S3 from some sort of an archive

pypt:
Fixes #274.

Plan:

* [x] Wait for frontend to drop Bit.ly support first
* [ ] Backup Bit.ly-related PostgreSQL data:
    * [x] `feeds.skip_bitly_processing` column
    * [x] `story_statistics_bitly_referrers` table
    * [x] `bitly_clicks_total` table
    * [x] `bitly_processing_schedule` table
    * [x] `snap.story_link_counts.bitly_click_count` column
    * [x] `snap.medium_link_counts.bitly_click_count` column
    * [x] `bitly_processing_results` table
* [ ] Deploy PR
* [ ] Remove Bit.ly-related PostgreSQL data listed above
* [ ] Backup S3 bucket with Bit.ly raw data
* [ ] Remove S3 bucket with Bit.ly raw data
* [ ] Remove Bit.ly configuration from `mediawords.yml`
* [ ] [Drop](https://lucene.apache.org/solr/guide/6_6/schema-api.html#SchemaAPI-DeleteaField) `bitly_click_count` Solr column on production
* [ ] Remove Bit.ly Supervisor logs
* [ ] Purge Bit.ly job queue on RabbitMQ
* [ ] Remove AWS credentials for Bit.ly object store


pypt:
> if we're going to delete the bitly fields from the snap.* tables, we need to archive that data off somewhere. those are big tables, so ideally we would archive just the bitly related fields plus the primary keys for the relevant tables.

I will both backup and delete every Bit.ly related column / table as listed in the to-do above.

> I don't see an update to the solr schema.xml in here.

Already done [here](https://github.com/berkmancenter/mediacloud/pull/322/files#diff-5cab371f6276048e0b3cb5f76041ec90).

> make sure we don't actually deploy this until the front end has been updated; otherwise, the front end will break in the various places it is expecting to be able to use the existing bitly-inclusive api

OK. Rahul, would you be able to drop Bit.ly support on your side first?

pypt:
Looks like https://github.com/mitmedialab/MediaCloud-Web-Tools/issues/894 has been finished. Backing up tables then.

PR #323, opened 2018-03-05
Reuse is_shortened_url() from mediawords.util.url

pypt:
We already have `is_shortened_url()` implemented in `mediawords.util.url`, so let's reuse that.

I've added `archive.is` and `feeds.feedburner.com` to the URL shortener list. Not sure why you had `twitter.com` there.


pypt:
> can you please add a separate check for twitter.com next to the is_shortened_url call?

Done in 115b6fa.

PR #344, opened 2018-03-22
change topics/create to public

ibrahimhaleemkhan:


pypt:
Hi Ibrahim, what's that?

hroberts:
Thanks for the pull request!

As a matter of policy, we require all significant use of our system to be through registered users.  We do this so that we can control load on the platform among other reasons.  That's a hard decision that we are unlikely to change.

Note that anyone can create a topic, so if you are submitting this request just to be able to create your own topic, you can already do that by just registering for an account.

Also, FYI, this PR would not serve its purpose because the topic creation code in various places assumes that the session is associated with a registered user, so it would just result in a crash instead of a permission error.

PR #368, opened 2018-05-04
Speed up sentence field count

### LINKED ISSUE #360: why are some field_count calls painfully slow?
I finally dug into something that has annoyed me - calls to `sentences/field_count` sometimes forever to return. This is particular true for tag sets 1963 (`nyt_labels`) and 1011 (`mc-geocoder`). See the attached CSV for details pulled from a recent topic mapper log file.  Calls for 1963 often took over 15 *minutes* to return; while those for 1011 routinely took a solid minute.  Any ideas why this would be?

[field_count_logs.csv.zip](https://github.com/berkmancenter/mediacloud/files/1934028/field_count_logs.csv.zip)

Notes to replicate generating the csv from a `server.log` file:
1. Run `cat server.log | grep "/field_count (with" > /tmp/field_count_logs.txt` to get the profiling log messages
2. Replace this disgusting regex: `.*Profiling: ([^ ]*)s .*timespans_id:([0-9]*).*sample_size': ([0-9]*).*q': '([^']*).*field': '([^']*).*tag_sets_id': ([0-9]*).*include_stats': ([0-9]).*` with `\1,\2,\3,\4,\5,\6,\7`
3. Add in the header row: `seconds,timespans_id,sample_size,q,field,tag_sets_id,include_stats`

pypt:
I've sped up the SQL query that does the tag matching -- apparently, fetching a list of tags for stories and filtering afterwards by tag set's ID is way faster (see ea81babc0e01991e7f1f367102c0bea42cb47c09).

Before: https://explain.depesz.com/s/6SAr -- 26.9 s; every `stories_tags_map` partition gets scanned a whopping 585,900 times (651 for all tags under tag set, times 900 for (probably) story ID count).

After: https://explain.depesz.com/s/x1Hx -- 0.5 s; every `stories_tags_map` partition gets queried for story IDs only; tag set filtering is done afterwards.

Also removed a bunch of dead / nonworking code and simplified parameters.

Fixes #360.

### DIFF
--- a/doc/api_2_0_spec/api_2_0_spec.md
+++ b/doc/api_2_0_spec/api_2_0_spec.md
@@ -849,12 +849,11 @@ are currently `tags_id_stories`.
 | `fq`            | `null`            | `fq` ("filter query") parameter which is passed directly to Solr
 | `sample_size`   | 1000              | number of sentences to sample, max 100,000
 | `include_stats` | 0                 | include stats about the request as a whole
-| `field`         | `tags_id_stories` | field to count
 | `tag_sets_id`   | `null`            | return only tags belonging to the given tag set
 
 See above /api/v2/stories_public/list for Solr query syntax.
 
-If the field is set to `tags_id_stories`, the call returns all of the tags associated with
+The call returns all of the tags associated with
 story including a sentence matching the query along with a count of how many times each tag is associated with
 each matching story.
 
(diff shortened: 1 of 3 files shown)


hroberts:
this looks good.

fyi, I am trying to replace all of this in the story_index branch with a much simpler module that just has solr do the tag counting through its faceting.

pypt:
Thanks for the info. I’ll merge it anyway as Rahul seems to be blocking on this.

PR #379, opened 2018-05-15
Partition "story_sentences"

pypt:
* Renamed a bunch of functions for them to be reused in future partitioning efforts:
    * `stories_partition_chunk_size()` returns number of stories to be stored in a single partition (up to 100m)
    * `stories_partition_name(base_table_name TEXT, stories_id INT)` returns a partition table name for a base table name, e.g. `stories_tags_map_01` for `stories_partition_name('stories_tags_map', 100000001)`
    * `stories_create_partitions(base_table_name TEXT)` creates missing partitions (plus one extra partition for future stories) for current `MAX(stories_id)` and returns a list of created partitions for further processing (e.g. for adding extra indexes)
    * All of the functions above are now used for partitioning `stories_tags_map` and `story_sentences`
* Partitioned `story_sentences` layout as discussed in #362:
    * Migration and the chunk copier automatically kills autovacuum on `story_sentences` before proceeding.
    * `story_sentences_id` are preserved between tables.
    * Didn't go for `md5(sentence)::uuid` because that's 128 bits (`half_md5()` returns only 64), also it's a bit too weird and confusing to store MD5 hashes as UUIDs (better to do it as `BYTEA`?)
    * Changing `stories_id` to `BIGINT` didn't work because a mismatch of `stories_id` type between the non-partitioned (`INT`) and partitioned (`BIGINT`) tables made the query planner opt for a sequential scan on either the non-partitioned table or all the partitions of the partitioned one. Played with various ways to cast stuff for an hour, didn't figure it out. I'd say let's change `stories_id` as a separate big step later on; we still have 50% of the available integer space, so there's no rush, and storing some `stories_id`s as `INT`s and others as `BIGINT`s might lead to more hassle (such as the hassle that I've just had by debugging this)
    * Let's remove `db_row_last_updated` as a separate task too: out of scope, too complex of a change, some more non-normalized columns would be nice to remove too. Don't get me wrong, I totally wanna do it, just as a separate task.
* Added a script `tools/db/copy_chunk_of_nonpartitioned_sentences_to_partitions.py` which, when run with Supervisor, copies 1m sentences from the non-partitioned table to the partitioned one.

Tested on mcdb2, works fine (tried upgrading schema, copying in chunks, querying, inserting, extracting, ...).

References #362.


pypt:
Thanks! Fixing some unit test bugs, will deploy on Monday as discussed over our meeting.

Issue #412, opened 2018-06-12
Clean up `scripts/` directory


pypt:
> scripts - fine to get rid of anything I don't explicitly list below
> * mediawords_fetch_url.pl - keep. I use this a couple times a year when I'm trying to debug why we can't download some feed or story. I could rewrite it, but it's annoying to have to spend the 15 minutes to rewrite it every time.
> * mediawrods_generate_daily_rss_dumps.pl - keep. this is what we are going to use to enable archive.org to mirror all of our content. I just updated and reran it a few weeks ago.
> * mediawords_generate_retweeter_foci / mediawords_generate_retweeter_scores - keep. this is what I use to do the retweeter score stuff (as with the retweet partisanship but also others).
> * mediawords_import_downloads_and_stories.pl - isn't this used by the backup crawler process to import the backup crawler content?
> * mediawords_query_to_regexp - I use this every once in a while to debug some issue with the query to regexp generation. saves me ten minutes of remembering how to do it in python.
> * mediawords_queue_jobs_from_query - This is fantastically useful a couple times a year when I need to queue up some set of jobs for the worker pool. It saves me from having to write a new one off script every time.
> * mediawords_run_tm_dump_query - keep. I use this every couple of months when I am trying to optimize a query that depends on the various temporary tables created during the snapshot creation or querying process.
> * mediawords_scrape_feedly - keep. I use this monthly to backfill some media sources.
> * mediawords_set_media_tags - keep. just added to mccore1 crontab in ansible.
> * mediawords_swap_solr_live_collection - probably keep. we in theory have the ability to switch between staging and production collections in solr. we have never actually used it in production, but I think we should keep this script around unless we want to refactor the rest of the solr code to remove this functionality.

pypt:
> mediawords_import_downloads_and_stories.pl - isn't this used by the backup crawler process to import the backup crawler content?

No, we use `./script/export_import/import_feed_downloads_to_db.pl` for that now:

https://github.com/berkmancenter/mediacloud-systems/blob/master/doc/backup_crawler.markdown#exporting

pypt:
Fixed in https://github.com/berkmancenter/mediacloud/pull/426 & deployed.

PR #426, opened 2018-06-19
Clean up old scripts under "script/"

pypt:
* Removed a bunch of scripts (in individual commits so that one could `git log --grep=...` and then `git revert` the script that's to be resurrected).

* Stripped the `mediawords_` prefix from scripts in `script/` for us to save those 200 ms (no need to type `med<TAB>`).


pypt:
Oh yeah, fixes #412.

PR #439, opened 2018-06-29
Drop Superglue support

### LINKED ISSUE #363: Drop Superglue support?
Superglue feeds are down again, so maybe drop its support after all?

@hroberts @rahulbot what do you think?

Related: https://github.com/berkmancenter/mediacloud-systems/issues/61

pypt:
Remove Superglue stories / sentences / feeds / downloads too.

(The rows are pre-deleted manually so deployment should be relatively fast.)

Fixes #363.


pypt:
Also, the deleted data is backed up to our "obsolete data" S3 bucket.

PR #477, opened 2018-09-25
Install and use Python 3.7

### LINKED ISSUE #295: Python 3.5 in security fix only mode
In case this is not on any radars, only source releases for security fixes from here until ~September 2020:

https://www.python.org/downloads/release/python-355/

Python 3.6 should continue to be supported until the start of 2022, and will enter a similar bugfix mode in ~July this year, after 3.7 is released (scheduled for June 15).

For what it is worth, I've been running parts of mediacloud on 3.6.4 with no problem.

pypt:
Install and use Python 3.7 because:

* It's the newest and greatest, has some fancy new features, e.g. dataclasses:

    * Dataclasses:

        ```python
        @dataclass
        class Point:
            x: float
            y: float
            z: float = 0.0

        p = Point(x=1.5, y=2.5)
        print(p)   # produces "Point(x=1.5, y=2.5, z=0.0)"
        ```

    * Perl-style f-strings:

        ```python
        >>> name = "Fred"
        >>> f"He said his name is {name}."
        'He said his name is Fred.'
        ```

    * Perl-style underscores in long numbers:

        ```python
        >>> 1_000_000_000_000_000
        1000000000000000
        ```

    * Variable typing support:

        ```python
        def foo():
            # The code will work just fine, but a smart IDE will complain
            # about `int` value being assigned to a variable that's
            # expected to be `str`
            name: str = 42
        ```

* It has became rather hard for me to maintain a working Python 3.5 binary on my Mac.

Also, update some dependency modules (including Readability) and `Inline::Python`.

More details:

* Updated [`readability-lxml`](https://pypi.org/project/readability-lxml/#history) from 0.6.2 (released in April 2016) to 0.7 (released in May 2018). The latest has some HTML5 improvements, e.g. ignores content in `<footer>`.
* Updated `Inline::Python` to the latest version by the [vendor](https://github.com/niner/inline-python-pm) plus Media Cloud's exception traceback changes. Original repo hasn't gone far, some floating point bugs have been fixed and that's pretty much it, so it should be safe to use.
* Made Ansible install Python 3.7 on Ubuntu 16.04 from [Dead Snakes](https://github.com/berkmancenter/mediacloud/compare/python_3_7?expand=1#diff-b370fd5907087514d81a09c0a9cfd608R167) repo, just like the rest of the internet does it.
* Replaced `sw-re2` with a [slightly more maintained version](https://github.com/berkmancenter/mediacloud/compare/python_3_7?expand=1#diff-f465cd9d2cab9385fc3e42cd3d29ad43R97).
    * Facebook's "official" [re2 library](https://github.com/facebook/pyre2/) is under development but doesn't support flags, e.g. `re.I`.
    * Most of the [forks](https://github.com/facebook/pyre2/network/members) are more or less abandoned, including `sw-re2`.
    * `andreasvc/pyre2` looks like it still sees some action, and it works on Python 3.7, however the updated version is not yet released so we install it from Git tree.
* Made `requirements.txt` install some modules from Git trees instead of released "stable" versions. Even though Python 3.7.0 was released this summer, some modules haven't caught up by releasing updated versions (there have been some breaking changes in Python's C API), and the 3.7 updates are present only in their development branches:
    * [flake8](https://github.com/berkmancenter/mediacloud/compare/python_3_7?expand=1#diff-f465cd9d2cab9385fc3e42cd3d29ad43R35), our code linter. No significant changes since the last stable, only some 3.7 improvements, and even if it breaks, it's no biggie.
    * [tweepy](https://github.com/berkmancenter/mediacloud/compare/python_3_7?expand=1#diff-f465cd9d2cab9385fc3e42cd3d29ad43R124) submitted some 3.7-related fixes recently.
* Renamed some packages, e.g. `.util.json` to `.util.parse_json` (`::Util::JSON` to `::Util::ParseJSON`), `.util.html` to `.util.parse_html` (same in Perl) because (probably) due to recent Python changes, the package names interfere with native `json` and `html` libraries provided by Python itself.
* Get rid of a bunch of warnings discovered by an IDE.

Fixes #295.


pypt:
Thanks for the review, will merge soon.

pypt:
Deployed.

PR #485, opened 2018-10-01
Generate and return C-compatible word2vec models, test loading older models

### LINKED ISSUE #484: gensim>=3.3.0 unable to load word2vec models generated with 3.2.0
After the Python 3.7 and dependencies upgrade (https://github.com/berkmancenter/mediacloud/pull/477), updated gensim==3.5.0 (or any other version) is unable to load word2vec models generated with gensim==3.2.0.

Reported to gensim here:

https://github.com/RaRe-Technologies/gensim/issues/2201

Going to remake our word2vec module to load / save ["C tool-compatible" models](https://radimrehurek.com/gensim/models/keyedvectors.html#gensim.models.keyedvectors.Word2VecKeyedVectors.save_word2vec_format) to improve compatibility.

Ping @rahulbot.

pypt:
* Due to gensim KeyedVectors incompatibilities between gensim==3.2.0 and newer versions, load and save word2vec models using the "C tool-compatible" format in the hopes that this is something that gensim will maintain compatibility with between versions.

* Add a test that attempts to load word2vec model generated with gensim==3.2.0.

Fixes #484.

@rahulbot, would you be able to replace:

```python
model = KeyedVectors.load(path_to_model_data)
```

to something like:

```python
try:
    # Try loading "new-style" C format model
    model = gensim.models.KeyedVectors.load_word2vec_format(path_to_model_data, binary=True) 
except Exception as ex:
    # Try loading "old-style" pickled model
    model = KeyedVectors.load(path_to_model_data)
```

on your end, and then let me know so that I could deploy this?


pypt:
It looks to me that snapshot's filename gets determined by `_topic_model_cache_name()`, and said function [doesn't append any kind of extension](https://github.com/mitmedialab/MediaCloud-WordEmbeddingsServer/blob/master/server/models.py#L26) so `_load_model_from_file()` will always try `gensim.models.keyedvectors.KeyedVectors.load()` instead of `gensim.models.keyedvectors.KeyedVectors.load_word2vec_format()`.

pypt:
Yeah, the idea is that you `try`-`except` during the interim period while we deploy our PR and update the word2vec models to the C-format on our end, and afterwards you could remove the `try`-`catch` block and just `load_word2vec_format()` everything.

pypt:
Deployed, added a task for Rahul to drop support for Pickled models: https://github.com/mitmedialab/MediaCloud-WordEmbeddingsServer/issues/9

PR #501, opened 2018-10-29
Cast API's integer parameters to int

pypt:
Not every integer API parameter was being correctly sanitized (e.g. `int($rows)`), and in some cases the code wasn't using database handler's parameter interpolation (`$db->query('SELECT * FROM foo WHERE bar = ?', $bar)`) but instead was using Perl's variable interpolation (`$db->query("SELECT * FROM foo WHERE bar = $bar")`) without any kind of escaping being done, thus making the code susceptible to SQL injections, e.g.:

```
$ curl "https://api.mediacloud.org/api/v2/tags/list?rows=%27;%20bobby%20tables&key=<...>"

{"error":"Error(s): Caught exception in MediaWords::Controller::Api::V2::Tags->list \"Traceback (most recent call last):\n  File \"/home/mediacloud/mediacloud/mediacloud/mediawords/db/result/result.py\", line 77, in __execute\n    cursor.execute(*query_args)\n  File \"/home/mediacloud/.virtualenvs/mediacloud/lib/python3.7/site-packages/psycopg2/extras.py\", line 144, in execute\n    return super(DictCursor, self).execute(query, vars)\npsycopg2.DataError: invalid input syntax for integer: \"'\"\nLINE 7:     order by tags_id asc limit ''''\n                                       ^\n\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/home/mediacloud/mediacloud/mediacloud/mediawords/db/handler.py\", line 309, in query\n    print_warnings=self.__print_warnings)\n  File \"/home/mediacloud/mediacloud/mediacloud/mediawords/db/result/result.py\", line 35, in __init__\n    print_warnings=print_warnings)\n  File \"/home/mediacloud/mediacloud/mediacloud/mediawords/db/result/result.py\", line 114, in __execute\n    'mogrified_query': str(mogrified_query),\nmediawords.db.exceptions.result.McDatabaseResultException: Query failed: invalid input syntax for integer: \"'\"\nLINE 7:     order by tags_id asc limit ''''\n                                       ^\n; query: ('select *\\n    from tags\\n    where\\n        tags_id > %s \\n        \\n        \\n    order by tags_id asc limit %s\\n', (0, \"'\")); mogrified query: b\"select *\\n    from tags\\n    where\\n        tags_id > 0 \\n        \\n        \\n    order by tags_id asc limit ''''\\n\"\""}
```

This wasn't necessarily exploitable but the `'; bobby tables` parameter reaching database level is bad enough IMHO.

I've cast every integer parameter that I could find to `int`, with default values set where applicable. I haven't checked whether there exist instances where we run a SQL query with unsanitized strings (e.g. Solr queries) though; for reference, those are to be sanitized with `$db->quote()`.

### DIFF
--- a/lib/MediaWords/Controller/Api/V2/Feeds.pm
+++ b/lib/MediaWords/Controller/Api/V2/Feeds.pm
@@ -137,7 +137,7 @@ sub scrape_status_GET
 {
     my ( $self, $c ) = @_;
 
-    my $media_id = $c->req->params->{ media_id };
+    my $media_id = int( $c->req->params->{ media_id } // 0 );
 
     my $job_class = MediaWords::Job::RescrapeMedia->name;
 
--- a/lib/MediaWords/Controller/Api/V2/MC_REST_SimpleObject.pm
+++ b/lib/MediaWords/Controller/Api/V2/MC_REST_SimpleObject.pm
@@ -127,9 +127,7 @@ sub _process_result_list
 
     if ( $self->has_nested_data() )
     {
-
-        my $nested_data = $c->req->param( 'nested_data' );
-        $nested_data //= 1;
+        my $nested_data = int( $c->req->param( 'nested_data' ) // 1 );
 
         if ( $nested_data )
         {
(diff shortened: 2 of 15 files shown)


hroberts:
this is great.

lots of these are '1' as a boolean, but there's no harm in casting those to ints as well.  

in some cases, this changes the behavior from only '1' triggers the boolean condition to 'any true perl scalar' triggers the boolean condition.  ideally we would keep the former definition just for consistency sake.

pypt:
Thanks! I see your point on the "any true Perl scalar" booleans, but I don't think Rahul & Cindy are sending us anything besides 1 / 0, so that's probably not that critical.

Issue #532, opened 2018-12-20
Downloads from the past have primary keys from the future


pypt:
![doc-brown-back-to-the-future](https://user-images.githubusercontent.com/181949/50304173-e5979480-0497-11e9-8a67-68cba736df12.png)

By sheer accident I have stumbled upon a download that has a very recent `downloads_id` but was fetched in ~2015:

```
mediacloud=# select * from downloads where downloads_id = 1810972751;
 downloads_id | feeds_id | stories_id | parent |                                               url                                               |      host       |       download_time        |  type   |      state      |      path       |                                                  error_message                                                  | priority | sequence | extracted 
--------------+----------+------------+--------+-------------------------------------------------------------------------------------------------+-----------------+----------------------------+---------+-----------------+-----------------+-----------------------------------------------------------------------------------------------------------------+----------+----------+-----------
   1810972751 |   435674 | 1119893954 |        | http://radaronline.com/features/2006/10/americas_dumbest_congressmen_a_radar_special_report.php | radaronline.com | 2015-04-06 02:08:43.046751 | content | extractor_error | content:pending | Died at /space/mediacloud/mediacloud_RELEASE_20140325/script/../lib/MediaWords/Util/ThriftExtractor.pm line 26.+|        1 |        1 | f
              |          |            |        |                                                                                                 |                 |                            |         |                 |                 |                                                                                                                 |          |          | 
(1 row)
```

By comparison, downloads "around" it (with `downloads_id` -1 or +1) were fetched just two days ago:

```
mediacloud=# select * from downloads where downloads_id in (1810972750, 1810972752) order by downloads_id;
 downloads_id | feeds_id | stories_id |   parent   |                      url                       |       host        |    download_time    |  type   |  state  |          path           | error_message | priority | sequence | extracted 
--------------+----------+------------+------------+------------------------------------------------+-------------------+---------------------+---------+---------+-------------------------+---------------+----------+----------+-----------
   1810972750 |  1101803 | 1119893956 | 1810906084 | https://www.al-madina.com/article/604281?rss=1 | www.al-madina.com | 2018-12-16 09:52:32 | content | success | s3:downloads/1810972750 |               |        0 |        1 | t
   1810972752 |  1101803 | 1119893957 | 1810906084 | https://www.al-madina.com/article/604280?rss=1 | www.al-madina.com | 2018-12-16 09:52:44 | content | success | s3:downloads/1810972752 |               |        0 |        1 | t
(2 rows)
```

[...]

Issue #530, opened 2018-12-18
Upgrade to PostgreSQL 11


pypt:
PostgreSQL 11 (released ~2 months ago), among other features, [has some nice partitioning improvements](https://www.postgresql.org/about/news/1894/) that I'd like to use for partitioning `downloads` and `download_texts` (#514):

* improved partition elimination strategy
* `UPDATE`s can now update the partition key and the row will automagically get moved to the partition where it belongs
* one can now create primary keys, foreign keys, indexes and triggers on master table, and the partitions will inherit them
* partitioned tables can now have `UNIQUE` indexes
* native atomic upsert now works on partitioned tables

It will also be nice to try out all the performance improvements, e.g. JIT query compilation.

As always, I'll do a test upgrade first and the actual production upgrade afterwards. Production upgrade will involve up to 1 hour of downtime.

**Plan for test upgrade:**

* [x] Install PostgreSQL 11 on Travis, see if it works (don't merge it yet).
* [x] Ensure that pre-`pg_upgrade` snapshot exists on backup mcdb2 to be able to restore to it later
* [x] Disable `zfs-send.sh` Cron job on production mcdb1.
* [x] Do a test `pg_upgrade` run on backup mcdb2, see how long does it take to run.
    * *(For 9.6, it took ~1 minute to run `pg_upgrade`.)*
    * *(For 9.6, it took ~10 minutes to run `vacuumdb --all ----analyze-only` to get the database to a usable state, another ~30 minutes to analyze all stages with `--analyze-in-stages`.)*
* [x] Come up with an estimate of how much planned downtime can we expect.
* [x] Reset backup mcdb2 to pre-`pg_upgrade` snapshot.
* [x] Reenable `zfs-send.sh` Cron job on production mcdb1.

**Plan for production upgrade:**

* [x] Back up production to Faith.
* [x] Set up date + time for upgrade, inform `mediacloud-team@`.
* [x] Disable `zfs-send.sh` Cron job on production mcdb1.
* [x] Run `zfs-send.sh` script to update snapshot on backup mcdb2 right before the upgrade.
* [x] Stop services, Apache, Solr imports on mccore1 / mcquery3.
* [x] Shut down PostgreSQL and PgBouncer on production mcdb1.
* [x] Run `zfs-send.sh` script to update snapshot on backup mcdb2.
* [x] Run `pg_upgrade` on production mcdb1.
* [x] Merge PostgreSQL 11 branch to `master` and `release`.
[...]

PR #533, opened 2018-12-20
Partition "downloads" and (partially) "download_texts"

### LINKED ISSUE #415: Remove old types from `download.types` enum and corresponding rows
`download_type`'s enum values `Calais`, `calais`, `spider_*`, `archival_only`, and relevant rows in `downloads` table

### LINKED ISSUE #514: partition downloads table
The downloads table is at >81% of its primary key space.  We need to partition it and change the key to bigint.

pypt:
Primary key of `downloads`, `downloads_id`, is running out of its `INT` range (currently at around 1,809,100,000 out of 2,147,483,647). So do the tables which reference `downloads` by `downloads_id`, e.g. `download_texts`. Additionally, `downloads` has grown quite big over the years (almost 2b rows!) and could use some partitioning to increase efficiency and make maintenance easier.

**Observations:**

* Row size for `downloads` is not that big but the row count is huge, we'll try to apply the generic "up to 100 GB and 100m rows per table" rule.
* Most performance-critical code runs queries against the `downloads` table with `type = '...'` and or `state = '...'` conditions, so we might benefit from further partitioning the table by `type` and `state` for the constraint exclusion to be able to kick in.
* This is a good chance to get rid of those obsolete rows (`'Calais'` and others).

**Statistics:**

About 1.8b rows in `downloads`:

```
mediacloud=# select max(downloads_id) from downloads;
    max     
------------
 1809105187 = 1,809,105,187
(1 row)
```

About 1.5b (not sure though, a good chunk of those could be deleted) rows in `download_texts`:

```
mediacloud=# select max(download_texts_id) from download_texts;
    max     
------------
 1532681794 = 1,532,681,794
(1 row)
```

Download counts per type and state:

```
mediacloud=# select type, state, count(downloads_id) from downloads group by type, state;
  type   |      state      |   count   
---------+-----------------+-----------
 Calais  | error           |      1935
 Calais  | success         |   2828267
 Calais  | extractor_error |         2
 calais  | error           |      1081
 calais  | success         |   4905065
 calais  | extractor_error |         5
 content | error           |  36242502
 content | fetching        |        45
 content | pending         |  66024205
 content | success         | 989372269
 content | extractor_error |    708611
 feed    | error           |  34166939
 feed    | fetching        |        13
 feed    | pending         |    367063
 feed    | success         | 614359087
 feed    | feed_error      |  16274800
 feed    | extractor_error |       368
(17 rows)
```

Table sizes:

```
mediacloud=# \d+
public | downloads      | table | mediacloud | 376 GB  |
public | download_texts | table | mediacloud | 1897 GB |
```

**`downloads` partitioning scheme:**

* `downloads_p` -- base table
    * `downloads_p_error` -- all rows with `state = 'error'` (~71m rows, ~16 GB)
    * `downloads_p_feed_error` -- all rows with `state = 'feed_error'` (~16m rows, ~4 GB)
    * `downloads_p_fetching` -- all rows with `state = 'fetching'` (around 100 rows at any given time)
    * `downloads_p_pending` -- all rows with `state = 'pending'` (~67m rows, 15 GB)
    * `downloads_p_success` -- base table for rows with `state =
      'success'`
        * `downloads_p_success_feed` -- base table for rows with `state = 'success' AND type = 'feed'`
            * `downloads_p_success_feed_00` -- rows with `downloads_id BETWEEN 1 AND 100,000,000` (i.e. up to 100m rows per partition) `AND state = 'success' AND type = 'feed'` (~35m rows and 8 GB per partition)
            * `downloads_p_success_feed_01` -- ...
        * `downloads_p_success_content` -- base table for rows with `state = 'success' AND type = 'content'`
            * `downloads_p_success_content_00` -- rows with `downloads_id BETWEEN 1 AND 100,000,000 AND state = 'success' AND type = 'content'` (~57m rows and 12 GB per partition)
            * `downloads_p_success_content_01` -- ...

**`download_texts` partitioning scheme:**

* `download_texts` -- base table
    * `download_texts_p_00` -- rows with `downloads_id BETWEEN 1 AND 100,000,000` (~57m rows and ~107 GB per partition)
    * `download_texts_p_01` -- ...

**Notes:**

* PostgreSQL 11 now has the ability to move rows between partitions on partitioning key change without using (slow) triggers and such, for example `UPDATE downloads SET state = 'success' WHERE state = 'fetching'` will get the row removed from `_pending` table and store it in one of the `_success` table's partitions in a single atomic operation
* As I understand from the crawler's design, downloads with `state = 'pending'` should be short-lived (once they get fetched, they should end up being of other `state`, e.g. `'success'`), but somehow we ended up having 67m of `'pending'` downloads. Is this a bug of some sort?
* Most of the churn will happen in `_pending` and `_fetching` tables (partitions), keeping their indices small and making it easier to maintain (e.g. autovacuum) them. All the rest of the partitions would then effectively serve as append-only, "write and forget" tables that shouldn't get much bloat at all.
* `type` and `state` enums would no longer have deprecated values in a partitioned table: https://github.com/berkmancenter/mediacloud/blob/partition_downloads/schema/mediawords.sql#L798-L826
    * `'[Cc]alais'` downloads wouldn't get migrated to the partitioned table.
    * Downloads with `state = 'queued'` would get moved as with `state = 'pending'`.
    * Downloads with `state = 'extractor_error'` would get moved as with `state = 'error'`, and their corresponding entries from `download_texts` would get removed. `'extractor_error'` state is weird because even though we have rows with their `state` set to `'extractor_error'` (and it kind of makes sense to have such a state), we don't *set* the download state to `'extractor_error'` anywhere in our codebase. Also, downloads with `state = 'extractor_error'` have their error message stored in a corresponding entry under `download_texts` which is a weird design (I hope they don't get extracted!).
* Partitioned tables still don't support foreign keys being pointed at them (the partitioned table can have foreign keys itself), so we try to maintain referential integrity with a bunch of triggers, some of which test whether an INSERTed / UPDATEd row exists in a referenced table, and others UPDATE / DELETE / set to NULL rows from referencing tables.
* `downloads_id` is now BIGINT everywhere and starts off where an unpartitioned table's `downloads_id` sequence left off.
* To switch from unpartitioned table to a partitioned one without downtime, just as with `story_sentences` we create an empty partitioned table, join the two tables using an updatable view (so that the application code could use `downloads` uninterrupted), and then gradually copy rows from the unpartitioned table to the partitioned one using a script.
* Even though we create partitioned table layouts for both `downloads` and `download_texts`, initially only rows from `downloads` are to be moved to a partitioned table. After partitioning `downloads`, `download_texts` could be made to reference a partitioned `downloads_p_success_content` natively (as we're only storing texts for successful content downloads in `download_texts`), and we could take our time copying `download_texts` rows afterwards.
* Other tables that reference `downloads` and need their `downloads_id` set to BIGINT either get DROPped and recreated (various cache tables the content of which is easily recreated), or simply copied over in a migration itself (`raw_downloads` which happens to be empty on production and has only a small number of rows in various "development" dumps that we have lying around).

**Still unsure about:**

* The sheer number of partitions. As per the proposed scheme, current dataset in `downloads` would get split up into 4 + 18 + 18 = 40 partition tables. In the most performance-dependend cases this is going to be advantageous as the query planner would get a chance to look for rows only in the specific partition based on query's `type` and `state`. However, queries with no such predicates (e.g. querying for a download based on it's `downloads_id` only) might encounter a performance hit as then the query planner wouldn't be able to employ constraint exclusion to limit the number of searched partitions. I'm not sure how significant this performance hit might be though because the executor is pretty fast at testing those 40 or so `downloads_id` indexes. I remember coming up with a large number of partitions while setting up Bit.ly click counts table, and I recall that it was the query planner that was the bottleneck of a typical query and not the executor, so maybe JIT (another feature of PostgreSQL) would be able to help us out with performance querying a large-ish number of partitions? With all that said, we can reduce the number of partitions from 40 to 4 + 18 = 22 by joining the `_success_feed` and `_success_content` sub-partitions into one.

**Misc. changes:**

* Stripped off `_partitioned` suffix from partitioned table, renaming it to just `_p`, e.g. `story_sentences_partitioned` is called `story_sentences_p` from now on. Less to type!
* Moved `cached_extractor_results` from `public` schema to `cache` and renamed it to `extractor_results_cache` (so now it's to be accessed as `cache.extractor_results_cache`), made it UNLOGGED, got rid of weird "do cleanup every X request" strategy and replaced it with a generic cleanup that we use for all of our tables in `cache`.

**Depends / fixes:**

Depends on #530 because we want PostgreSQL 11's ability to move rows between partitions on partition key change.

Probably depends on #532 too as I'm not sure what's happening with `downloads_id` sequence.

Fixes #514.

Also fixes #415 as we skip copying rows with obsolete download types.


hroberts:
Comments on the plan, before looking at the code:

> * As I understand from the crawler's design, downloads with `state = 'pending'` should be short-lived (once they get fetched, they should end up being of other `state`, e.g. `'success'`), but somehow we ended up having 67m of `'pending'` downloads. Is this a bug of some sort?

There are a couple reasons for long lived pending downloads:
* we throttle downloads to one request per $engine->throttle() seconds per host.  57.5 million of the pending downloads are for the most common 100 hosts.   I think the existing throttle() setting (set in script/crawl.py) of 10 was just not letting those hosts catch up.  I have decreased the throttle() setting to 1 so that we can hopefully start catching up on those downloads.
* we also retry downloads that fail with a 500 or 503 error code up to 10 times by setting the state of the download to 'pending' but setting the download_time to $number_of_retry hours in the future.

>   * Downloads with `state = 'extractor_error'` would get moved as with `state = 'error'`, and their corresponding entries from `download_texts` would get removed. `'extractor_error'` state is weird because even though we have rows with their `state` set to `'extractor_error'` (and it kind of makes sense to have such a state), we don't _set_ the download state to `'extractor_error'` anywhere in our codebase. Also, downloads with `state = 'extractor_error'` have their error message stored in a corresponding entry under `download_texts` which is a weird design (I hope they don't get extracted!).

At some point, we were setting the state to 'extractor_error' when there was an extraction error.  The idea was to record that there was a problem with the extraction but still allow fetching of the download content.  If the state of a download is 'error', the fetch_content() function will return an error even if there is content associated with the download present in the store.  From staring at the code, it looks like now we just leave the state of a download with a failed extraction as 'success'.  I think storing the extraction error in the downloads table (and especially the state field) is crappy, but we should be storing those failures somewhere.  Regardless, we should not set the existing 'extractor_error' states to 'error', because that will disable content fetching for those downloads.    Maybe just create a 'download_extractor_errors' table, record the existing and future extractor errors there, and then set the 'extractor_error' states to 'success'?

> * Partitioned tables still don't support foreign keys being pointed at them (the partitioned table can have foreign keys itself), so we try to maintain referential integrity with a bunch of triggers, some of which test whether an INSERTed / UPDATEd row exists in a referenced table, and others UPDATE / DELETE / set to NULL rows from referencing tables.

ugh. seems just to be a cost of using postgres partitioning.

> * `downloads_id` is now BIGINT everywhere and starts off where an unpartitioned table's `downloads_id` sequence left off.

yes

> * To switch from unpartitioned table to a partitioned one without downtime, just as with `story_sentences` we create an empty partitioned table, join the two tables using an updatable view (so that the application code could use `downloads` uninterrupted), and then gradually copy rows from the unpartitioned table to the partitioned one using a script.

yes

> * Even though we create partitioned table layouts for both `downloads` and `download_texts`, initially only rows from `downloads` are to be moved to a partitioned table. After partitioning `downloads`, `download_texts` could be made to reference a partitioned `downloads_p_success_content` natively (as we're only storing texts for successful content downloads in `download_texts`), and we could take our time copying `download_texts` rows afterwards.

yes

> * Other tables that reference `downloads` and need their `downloads_id` set to BIGINT either get DROPped and recreated (various cache tables the content of which is easily recreated), or simply copied over in a migration itself (`raw_downloads` which happens to be empty on production and has only a small number of rows in various "development" dumps that we have lying around).

yes

> **Still unsure about:**
> 
> * The sheer number of partitions. As per the proposed scheme, current dataset in `downloads` would get split up into 4 + 18 + 18 = 40 partition tables. In the most performance-dependend cases this is going to be advantageous as the query planner would get a chance to look for rows only in the specific partition based on query's `type` and `state`. However, queries with no such predicates (e.g. querying for a download based on it's `downloads_id` only) might encounter a performance hit as then the query planner wouldn't be able to employ constraint exclusion to limit the number of searched partitions. I'm not sure how significant this performance hit might be though because the executor is pretty fast at testing those 40 or so `downloads_id` indexes. I remember coming up with a large number of partitions while setting up Bit.ly click counts table, and I recall that it was the query planner that was the bottleneck of a typical query and not the executor, so maybe JIT (another feature of PostgreSQL) would be able to help us out with performance querying a large-ish number of partitions? With all that said, we can reduce the number of partitions from 40 to 4 + 18 = 22 by joining the `_success_feed` and `_success_content` sub-partitions into one.

I think 40 partitions should be fine.  I think this starts to be a problem hundreds or thousands of partitions.

FWIW, the queries that have caused the most problems after partitioning have been big join queries, such as joins of stories to stories_tags_map.  I looked through the code and didn't find many performance critical downloads joins.  We just don't use this table that much once the processing is done.

> 
> **Misc. changes:**
> 
> * Stripped off `_partitioned` suffix from partitioned table, renaming it to just `_p`, e.g. `story_sentences_partitioned` is called `story_sentences_p` from now on. Less to type!
> * Moved `cached_extractor_results` from `public` schema to `cache` and renamed it to `extractor_results_cache` (so now it's to be accessed as `cache.extractor_results_cache`), made it UNLOGGED, got rid of weird "do cleanup every X request" strategy and replaced it with a generic cleanup that we use for all of our tables in `cache`.

sounds good.

ansible/roles/postgresql-server/tasks/install-Ubuntu.yml
```diff
@@ -56,6 +56,11 @@
     - "postgresql-contrib-9.6"
     - "postgresql-plperl-9.6"
     - "postgresql-server-dev-9.6"
+    - "postgresql-10"
+    - "postgresql-client-10"
```
hroberts: should this be postgresql-11?

pypt: That's the Ansible task that removes old versions of PostgreSQL before installing the designated version (11 in this case).

pypt:
Thanks for the review! Very useful to have another set of eyes on such a risky change.

> Maybe just create a 'download_extractor_errors' table, record the existing and future extractor errors there, and then set the 'extractor_error' states to 'success'?

Reimplementing `extractor_error` is a new code change that has to be made and tested, so how about if I copy `downloads_id` with `state = 'extractor_error'` to a separate table for now, deploy the PR as per the original plan, and then we can resolve the `extractor_error` stuff in the future as a separate task?

hroberts:
great
.

Issue #585, opened 2019-05-15
add support for new associated press api


hroberts:
We just renewed access to the associated press feed, but they are no longer giving access to their old, rss based feed.  Instead, we have to ingest their custom api.

We should build this as a python module mediawords.crawler.download.feed.ap that has a single function called get_new_stories() or similar that goes to the api, downloads the list of the latest 100 stories, then downloads the content of each of those stories, then returns a list of the stories as a list of dicts.  

This task only needs to do the work to ingest and format the data.  We have a somewhat complicated system on the perl side that does all of the crawler scheduling and integration.  We will assign as a separate task the little bit of work to write the little bit of perl glue code that will do the remaining work to stick the resulting data into the database.

The python module should return list of dicts with the following fields: url, publish_date, title, description, text.  After ten minutes of poking at the api docs, I think we will have to first call the /feeds end point to get the list of most recent stories and then fetch a separate url for each story to get the actual content for each story.

For rss feeds, the crawler works on a model of fetching the rss feed as one job and then inserting any new stories as new job downloads.  That would be a pain in this case because the ap content downloads would require code to handle api authentication and parsing of the custom ap content format.  I think it will be much simpler to have the single get_new_stories() function just download each of the new pieces of content in serial.  We should only get a couple hundred stories a day from the ap, so it should be quick to just download the handful we might get for a given feed download.

One problem we ran into with the previous ap integration is that the ap constantly updates its stories for the first few hours after they are published.  Our system is not built to handle constant changes to stories (or to have different versions of stories).  The old rss feed based system handled this problem by waiting 12 hours before actually downloading any of the urls found in the ap rss feed.  We can't do that if we are downloading stories as we discover them, but maybe we can just ask the api for a feed of stories that are 12 hours old? 

I will send the api instructions and authentication separately.

pushshift:
Thanks. I'm reviewing the API documentation now. Data ingest is my specialty. :)

When you import the module and instantiate an instance, I'm assuming it would make sense for us to put some parameters that can be passed such as one for the API key itself, max/min story time, etc. I'll continue reading up on the API docs and incorporate your specs into the module.

hroberts:
You should put the api key into the configuration file (mediawords.yml) and access it via mediawords.util.config.get_config().

You should just return all of the stories each time.  I will deal with deduping the stories with the existing ap stories in the database on the perl side.

[...]

PR #588, opened 2019-05-17
Pull Request for https://github.com/berkmancenter/mediacloud/issues/585

pushshift:


mediacloud/mediawords/crawler/download/feed/ap.py
```diff
@@ -0,0 +1,207 @@
...
+        params.update(kwargs)
+        content_data = self._make_request(url,params)
+        return content_data
+
+    def _make_request(self,
+                      url: str,
+                      params: dict = None) -> str:
+        """Internal method for making API requests"""
+
+        retries = self.retry_limit
```
hroberts: we have a weird homebrew implementation of LWP in python so that our web fetching code in perl and python goes through the same code path.  we include support for retries in that code.  that code makes sure to set the user agent and from headers to the configured values and also has some media cloud custom hooks, such as logging every web request we make.

here's some example code:

```
import mediawords.web.user_agent

ua = mediawords.web.user_agent.UserAgent()
ua.set_timing([1, 2, 4, 8, 16, 32, 64, 128, 256])
response = ua.get(url)
if response.is_success:
    return response.decoded_content()
else:
    raise SomeError("Unable to fetch url '%s': %s" % (url, response.as_string()))
```

mediacloud/mediawords/crawler/download/feed/ap.py
```diff
@@ -0,0 +1,207 @@
...
+                    logger.debug("Successfully retrieved {}".format(url))
+                    return response.content
+                else:
+                    logger.warning("Received HTTP status code {} when fetching {}".format(response.status_code,url))
+
+            retries -= 1
+
+            if retries == 0:
+                logger.error("Could not fetch {} after {} attempts. Giving up.".format(url,self.retry_limit))
+                raise ValueError("Could not fetch {} after {} attempts. Giving up.".format(url,self.retry_limit))
```
hroberts: ideally we would use a custom error class here such as 'McAPFetchError' or whatever

mediacloud/mediawords/crawler/download/feed/ap.py
```diff
@@ -0,0 +1,207 @@
...
+
+            wait_time = (self.retry_limit - retries) ** 2
+            logger.info("Exponentially backing off for {} seconds.".format(wait_time))
+            time.sleep(wait_time)
+
+
+def get_new_stories(db=None) -> list:
```
hroberts: the db should be type hinted to mediawords.db.DatabaseHandler

mediacloud/mediawords/crawler/download/feed/ap.py
```diff
@@ -0,0 +1,207 @@
...
+
+        item = obj['item']
+        guid = item['altids']['itemid']
+        version = item['version']
+
+        # If DB handle was passed, check if this story has previously been retrieved (to avoid unnecessary API calls to content endpoint)
+        if db:
+            guid_exists = db.query("select 1 from stories s join media m using (media_id) where m.name = 'AP' and s.guid = %(a)s", {'a': guid}).hash()
+            if guid_exists:
+                logger.info('Story with guid: {} is already in the database -- skipping.')
```
hroberts: I would make this logger.debug().  my guess is that in production 98 of the 100 stories are going to be existing stories, so this will get very verbose.

mediacloud/mediawords/crawler/download/feed/ap.py
```diff
@@ -0,0 +1,204 @@
...
+import json
+import os
+import time
+import logging
+import urllib.parse as urlparse
+from typing import Any
+from bs4 import BeautifulSoup
+from mediawords.util.config import get_config
+
+logging.basicConfig()
```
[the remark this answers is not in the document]
pushshift: Will do -- thanks!

mediacloud/mediawords/crawler/download/feed/test_ap.py
```diff
@@ -0,0 +1,95 @@
...
+        httpretty.reset()
+
+
+    def test_fetch_nitf_rendition(self) -> None:
+        """Test fetching of nitf content and that it is valid XML"""
+        story_item = self.fixture_data_stories[0]['item']
+        nitf_content = ap._fetch_nitf_rendition(story_item)
+        soup = BeautifulSoup(nitf_content,features="html.parser")
+        body_content = soup.find('body.content').text
+        assert body_content
```
hroberts: I would assert something a little more specific about the body content.  if you can't determine the exact body content, at least assert its length.  it is really easy to imagine an error in which the tested function is failing but returning something trivial that passes this test.

PR #635, opened 2019-12-02
Use CTE instead of temporary tables to query for pending downloads

pypt:
Temporary tables fill up "pg_class" and "pg_attribute" so let's reduce those.

As of PostgreSQL 11, all CTEs get materialized first, thus making CTE a safe replacement for a temporary table.

Replaced the recursive trick with LATERAL join to get a single download for each parameter host.

Removed the "queued_downloads" copying as it's unclear how a copy could have prevented issues with dead rows. Query plans for the updated query seem to be choosing sequential scans when the number of rows in "queued_downloads" is low and index scans with a bigger number of rows.

Sample query plan:

```
                                                                                                       QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=73.85..152.91 rows=100 width=8) (actual time=207.795..833.113 rows=5 loops=1)
   CTE pending_hosts
     ->  CTE Scan on t t_1  (cost=71.27..73.29 rows=100 width=32) (actual time=197.140..792.443 rows=5 loops=1)
           Filter: (host IS NOT NULL)
           Rows Removed by Filter: 1
           CTE t
             ->  Recursive Union  (cost=0.56..71.27 rows=101 width=32) (actual time=197.137..792.430 rows=6 loops=1)
                   ->  Limit  (cost=0.56..0.66 rows=1 width=21) (actual time=197.133..197.134 rows=1 loops=1)
                         ->  Index Only Scan using downloads_pending_host_priority_downloads_id on downloads_pending downloads_pending_2  (cost=0.56..868758.95 rows=8471159 width=21) (actual time=197.132..197.132 rows=1 loops=1)
                               Heap Fetches: 93093
                   ->  WorkTable Scan on t  (cost=0.00..6.86 rows=10 width=32) (actual time=99.211..99.211 rows=1 loops=6)
                         Filter: (host IS NOT NULL)
                         Rows Removed by Filter: 0
                         SubPlan 1
                           ->  Limit  (cost=0.56..0.67 rows=1 width=21) (actual time=119.045..119.046 rows=1 loops=5)
                                 ->  Index Only Scan using downloads_pending_host_priority_downloads_id on downloads_pending downloads_pending_1  (cost=0.56..298862.22 rows=2823720 width=21) (actual time=119.043..119.043 rows=1 loops=5)
                                       Index Cond: (host > t.host)
                                       Heap Fetches: 601389
   ->  CTE Scan on pending_hosts  (cost=0.00..2.00 rows=100 width=32) (actual time=197.150..792.459 rows=5 loops=1)
   ->  Limit  (cost=0.56..0.76 rows=1 width=31) (actual time=8.126..8.128 rows=1 loops=5)
         ->  Nested Loop Anti Join  (cost=0.56..277095.31 rows=1411859 width=31) (actual time=8.125..8.125 rows=1 loops=5)
               Join Filter: (queued_downloads.downloads_id = downloads_pending.downloads_id)
               ->  Index Only Scan using downloads_pending_host_priority_downloads_id on downloads_pending  (cost=0.56..149876.83 rows=1411860 width=31) (actual time=7.623..7.623 rows=1 loops=5)
                     Index Cond: (host = pending_hosts.host)
                     Heap Fetches: 72441
               ->  Materialize  (cost=0.00..151.09 rows=6 width=8) (actual time=0.499..0.499 rows=0 loops=5)
                     ->  Seq Scan on queued_downloads  (cost=0.00..151.06 rows=6 width=8) (actual time=2.483..2.483 rows=0 loops=1)
 Planning Time: 2.722 ms
 Execution Time: 833.250 ms
(29 rows)
```

Runs in a comparable speed or faster than the previous implementation with temporary tables.

Lastly, add an undocumented index from production to the repository schema.

### DIFF
--- a/apps/crawler-provider/src/python/crawler_provider/__init__.py
+++ b/apps/crawler-provider/src/python/crawler_provider/__init__.py
@@ -160,9 +160,46 @@ def provide_download_ids(db: DatabaseHandler) -> None:
 
     log.info("querying pending downloads ...")
 
-    # get one downloads_id per host, ordered by priority asc, downloads_id desc, do this through a plpgsql
-    # function because that's the only way to avoid an index scan of the entire (host, priority, downloads_id) index
-    downloads_ids = db.query("select get_downloads_for_queue() downloads_id").flat()
+    # get one downloads_id per host, ordered by priority asc, downloads_id desc
+    downloads_ids = db.query("""
+
+        -- This "WITH RECURSIVE" trick works faster than "SELECT DISTINCT host
+        -- FROM downloads_pending" with lots of dead rows in the table
+        WITH pending_hosts AS (
+            WITH RECURSIVE t AS (
+                (SELECT host FROM downloads_pending ORDER BY host LIMIT 1)
+                UNION ALL
+                SELECT (SELECT host FROM downloads_pending WHERE host > t.host ORDER BY host LIMIT 1)
+                FROM t
+                WHERE t.host IS NOT NULL
+            )
+            SELECT host FROM t WHERE host IS NOT NULL
+        )
+
+        SELECT dp.downloads_id
+        FROM pending_hosts,
+
+        -- Get a top download for every host
+        LATERAL (
+            SELECT downloads_id
+            FROM downloads_pending
+            WHERE downloads_pending.host = pending_hosts.host
+              AND NOT EXISTS (
+
+                -- Not queued yet
+                SELECT 1
+                FROM queued_downloads
+                WHERE queued_downloads.downloads_id = downloads_pending.downloads_id
+
+              )
+            ORDER BY
+                downloads_pending.host,
+                priority,
+                downloads_id DESC NULLS LAST
+            LIMIT 1
+        ) AS dp
+
+    """).flat()
 
     log.info("provide downloads host downloads: %d" % len(downloads_ids))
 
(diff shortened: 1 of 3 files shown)


hroberts:
I didn't actually know about lateral joins.  very nice!

the issue with queued_downloads is that it will often have a few tens or hundreds of thousands of dead rows.  I have the autovacuum settings cranked up to vacuum that table aggressively, but it churns through millions of downloads a day, so it ends up as many as a million dead rows regardless.  At times, we have as many as 10k media sources that each have at least one row in downloads_pending.  Both my original code and your below code basically act as loops querying one download per media source and then testing to make sure that download does not already exist in queued_downloads.  If that bloated queued_downloads table takes just a second to query, the larger looping query will take 10k seconds to query.  By just copying queued_downloads into a temp table, we avoid querying a bloated table thousands of times in a row.  This was a fix for an actual performance problem encountered in production -- it's not just a theoretical optimization.  The way to test for it is to generate a queued_downloads table with 1 million dead rows and then a downloads_pending table with 10k distinct media sources and then see if your query still performs well.

did you test your new query with ~100 million rows in the downloads_pending table?  we have gotten into trouble assuming that we can chug along with a small queue in the past and then having things explode once we get above a couple million downloads pending, so I try hard to make sure these queries still work with a very large downloads_pending queue.  sometimes the downloads_pending queue gets big just because we fall behind for whatever reason, but it can also happen when we add a big chunk of new sources or backfill a bunch of sources from feedly.

pypt:
Solved in https://github.com/mediacloud/backend/pull/791

PR #636, opened 2019-12-02
Remove "_u" temporary table in find_dup_story()

pypt:
* Super hard to find due to the name
* Spams the "pg_catalog" tables
* Could be easily redone into a CTE


hroberts:
looks good.  we just need to keep an eye on the running queries for a few days after releasing this to make sure it is not exploding any query plans.  I think I have had the experience in the past of CTEs performing differently than temp tables because the query planner seems less aware of how many results are in a CTE resulty.

Issue #639, opened 2019-12-11
3rd party service monitoring


pypt:
> Now that we consume 3rd party APIs as part of our core pipeline, we need some system for monitoring them to find out when they are down. This should be a sanity test to make sure the APIs we consume are still working, and email us or alert us in some way when they aren't.
>
> @pypt suggested a Munin plugin could work for this. Should we queue this up as a dev task on the main MC repo, or is there some other solution that would be better?

pypt:
I now think that having a separate Munin plugin that would monitor something like Facebook Graph API would not the most optimal solution because 1) we would have to come up with and maintain a separate implementation of the same 3rd party API client code in the plugin; 2) the plugin can't possibly test for all of the border cases.

I thereby would like to propose a possibly better approach: **distinguishing between soft and hard failures, and quitting the process on the latter**.

In any part of our processing chain, two types of errors can occur:

* **Soft errors** happen when something goes wrong but we can reasonably expect that we can continue operation. Typical soft errors are usually faulty inputs: bad URLs, nonexistent stories / downloads, webpage fetch errors, etc. On soft errors, we still have to report failure back to the worker script (in the form of exception) but also we should continue running whatever it is that we're running (e.g. by failing a fetch operation for a particular story and moving on to a new one).

* **Hard errors** happen when something goes wrong and it doesn't look like we can continue operating with current or any future inputs. Hard errors usually occur when assertions are not met or when a program encounters something that it (or the programmer) hasn't seen before, e.g. HTTP client unable to decode JSON on requests when response is expected to be JSON in all cases, `list` parameter instead of an expected `dict`, required configuration parameters not being set, or something along those lines.

Right now, we treat both kinds of errors the same: we throw them as the same type of generic exception, and then the caller either catches them, or lets the process crash. I propose that after segregating those two kinds of errors (by throwing two different types of exceptions), we treat them differently too:

* On soft errors, we should **log the error but continue operation with subsequent inputs** as we can reasonably expect for the failure to not be permanent and mess up other inputs too. For example, if a particular story's URL can't be parsed, that doesn't mean that the next story will fail too; if a story that's being processed doesn't exist, maybe another one will be present; if a webpage fetch fails, then we can just try the next URL.

[...]

PR #640, opened 2019-12-11
Rewrite Facebook stats worker to Python, improve error reporting

pypt:
While in operation, the Facebook API client (`facebook_fetch_story_stats.*` module) differentiates between "soft" and "hard" errors, and throws two different types of `McFacebookException` subclasses:

* `McFacebookSoftFailureException` -- a "soft" error which means that the API call failed for those particular inputs for whatever reason, but that the worker can continue making API calls for other inputs.

* `McFacebookHardFailureException` -- a "hard" error which occurs on unexpected responses, permanent API errors, etc. and on which the worker should stop operation as the subsequent requests are not expected to be successful.

The worker, in turn:

* On a "soft" exception (`McFacebookSoftFailureException`), reraises the exception to fail a particular job (for a particular story) and continues with other jobs in the queue

* On a "hard" exception (`McFacebookHardFailureException`) or any other exception (which is assumed to be an unknown and thus hard failure), reports the exception to STDERR and exits the worker process with a non-zero error code (with `fatal_error()`) for the worker to shut down as it's assumed that upon a hard failure, the worker can't be productive anymore and someone needs to take a manual look into it

The `facebook-fetch-story-stats` app in production's `docker-compose.yml` is configured to auto-restart a failed worker up to three times and give up afterwards. This way a developer can learn about a failing app from Docker Swarm or Portainer.

Also:

* Reintroduce retries on retryable (potentially temporary) errors
* Store reaction count into `story_statistics.facebook_reaction_count`

References #639.


hroberts:
I'm good with merging this.  I think I already deployed a kludgier solution that just refuses to let the facebook fetcher die.  I think we will have to adjust this as we go to avoid dying too often as fb returns bad responses, but the right approach is to make those adjustments as needed.

pypt:
Thanks for the review on this PR and others! I think I'll deploy all three PRs on Monday, don't want to ruin my evening with some potentially botched code.

pypt:
Deployed, nothing's in the Facebook fetch queue so not sure how it will work.

PR #653, opened 2020-01-08
Rewrite feed parsing to Python

pypt:
Rewrite RSS / Atom feed parsing to Python, use [`feedparser`](https://pypi.org/project/feedparser/) behind the scenes.

`feedparser` is a mature library with a lot of traction and well developed features. My favorite thing about the module is that it uses SGML parsing instead of XML, so it tolerates a lot of feed errors that we might encounter in the wild and is supposedly not vulnerable to all of those [XML hacks](https://pypi.org/project/defusedxml/).

In `parse_feed()`, I've removed a bunch of `XML::FeedPP` tricks (removing comments etc.) as they're probably not relevant to the `feedparser`, and eliminated that huge `_fix_atom_content_element_encoding()` hack as I'm not sure what it was doing anyway, there weren't any tests covering that functionality, and `feedparser` [should be able to do its own sophisticated content encoding detection](https://pythonhosted.org/feedparser/character-encoding.html).

In the Perl implementation, feed item's `description()` would sometimes skip a possibly lengthier and more extensible `<content>` or `<content:encoded>` of the item and fall back to `<description>`. That is no longer the case - if `<content>` exists, `description()` will return that and skip `<description>` altogether.

In the Perl implementation, `MediaWords::DBI::Media::add_feed_url_to_medium()` would call `XML::FeedPP` directly that would both fetch the feed and try to parse it. I've replaced it with a tandem of `MediaWords::Util::Web::UserAgent()` + `MediaWords::Feed::Parse::parse_feed()` so that it gets a chance to make use of all the HTTP fetching tricks that we have implemented in `UserAgent` class, there's now a timeout of how long the feed will be attempted to be fetched, and `parse_feed()` will be way more forgiving with its inputs thanks to the underlying `feedparser`.

Other than that, it's a typical Perl -> Python rewrite with the Perl proxy class and all.


hroberts:
sorry for being a typical pita, but can we test this against the current version to find cases where the current works but this breaks?  my experience with rss is that it is generally horribly broken, and different parsers to varying jobs of dealing with various kinds of brokenness.  my intuition is that the newer python module will do a better job overall, but I would like to minimize surprises.

pypt:
How do I do this? Fetch a bunch of RSS feeds and try parsing them?

hroberts:
first pass would be:

* query db for all feed downloads that have at least one story for a given
day
* for each one, run the old feed parser and new feed parser
* compare: number of feed parsing errors, number of stories returned for
each feed, length of content/description, length of title, and publish date
for each story

maybe compare other data if you look at it and it makes sense.  I wouldn't
go crazy trying to figure out how to match up the data fields to be able to
do exact comparisons of field values, but we should get a sense of how
often each approach encounters errors, whether we are getting the same
number of stories, and some sanity tests to make sure we are getting
reasonable values back for description and title.

-hal

pypt:
I couldn't think of an easy way to "query db for all feed downloads that have at least one story for a given day", mostly because most of the feed downloads are those "redundant feeds", and we can figure out whether it's "redundant" or not only by querying S3.

So, instead I got the URLs of all the feeds except for the weird ones:

```sql
SELECT feeds_id, url
FROM feeds
WHERE type = 'syndicated'
  AND url NOT LIKE '%#spiderfeed'
  AND url NOT LIKE 'MediaWords::ImportStories::Feedly%'
  AND url NOT LIKE 'feed://%'
  AND active = 't'
ORDER BY feeds_id
;
```

Fetched current versions of all of them:

```python
#!/usr/bin/env python3
#
# cat feeds.csv | ./print_curl_commands.py | shuf > curl_commands.txt
# cat curl_commands.txt | parallel -j 128
#

import csv
import os
import shlex
import sys

csv_reader = csv.reader(sys.stdin, delimiter=',')
for row in csv_reader:
    feeds_id = row[0]
    url = row[1].strip()

    command = f'if [ ! -s feeds/{feeds_id}.xml ]; then curl --silent --max-time 10 --output feeds/{feeds_id}.xml {shlex.quote(url)} && echo {feeds_id}; fi'
    print(command)
```

Then wrote scripts for both [old feed parser in Perl](https://gist.github.com/pypt/98a3bf83be5da11f9cbe0a9ec3a6e83d) and [new feed parser in Python](https://gist.github.com/pypt/915c0fd5bf4258bee584221bfb5514db) which would parse the feed and then print the following:

* `parse_succeeded` - whether or not an attempt to parse the feed succeeded;
* `item_count` - how many items were found in a feed;
* `feed_title_length` - length of feed's (not the item's) `<title>`;
* `total_items_title_length` - total length of all feed item `<title>`s;
* `total_items_description_length` - total length of all feed item `<description>`s (or `<summary>`ies in Atom);
* `total_items_defined_publish_dates` - number of items that had their publication date defined and successfully parsed by the parser;
* `average_item_timestamp` - average publication date UNIX timestamp of all items in a feed.

I then ran both scripts by mounting them into an appropriate container image:

```shell
# Old (Perl) feed parser
docker pull dockermediacloud/common:release
mkdir feed_results_old/
docker run -it \
    -v "$(pwd)"/feeds/:/feeds \
    -v "$(pwd)"/feed_results_old/:/feed_results_old \
    -v "$(pwd)"/validate_old.pl:/validate_old.pl \
    dockermediacloud/common:release \
    perl /validate_old.pl

# New (Python) feed parser
docker pull dockermediacloud/common:feed_parser_python
mkdir feed_results_new/
docker run -it \
    -v "$(pwd)"/feeds/:/feeds \
    -v "$(pwd)"/feed_results_new/:/feed_results_new \
    -v "$(pwd)"/validate_new.py:/validate_new.py \
    dockermediacloud/common:feed_parser_python \
    python3.7 /validate_new.py
```

Merged the two resulting CSVs into one and imported everything into the following schema:

```sql
CREATE TABLE scratch.feed_parser_comparison (
    feeds_id    INT NOT NULL,

    old_parse_succeeded                     BOOL NOT NULL,
    old_feed_title_length                   INT NULL,
    old_item_count                          INT NULL,
    old_total_items_title_length            INT NULL,
    old_total_items_description_length      INT NULL,
    old_total_items_defined_publish_dates   INT NULL,
    old_average_item_timestamp              BIGINT NULL,

    new_parse_succeeded                     BOOL NOT NULL,
    new_feed_title_length                   INT NULL,
    new_item_count                          INT NULL,
    new_total_items_title_length            INT NULL,
    new_total_items_description_length      INT NULL,
    new_total_items_defined_publish_dates   INT NULL,
    new_average_item_timestamp              BIGINT NULL
);

\copy scratch.feed_parser_comparison FROM 'feed_parser_comparison.csv' DELIMITER E'\t' CSV HEADER;

CREATE UNIQUE INDEX scratch_feed_parser_comparison_feeds_id
    ON scratch.feed_parser_comparison (feeds_id);

CREATE INDEX scratch_feed_parser_comparison_parse_succeeded
    ON scratch.feed_parser_comparison (old_parse_succeeded, new_parse_succeeded);
```

So then, some super lazy stats:

```sql
# SELECT
    (
        SELECT COUNT(*)
        FROM scratch.feed_parser_comparison
        WHERE old_parse_succeeded
    ) AS total_old_parse_succeeded,
    (
        SELECT COUNT(*)
        FROM scratch.feed_parser_comparison
        WHERE new_parse_succeeded
    ) AS total_new_parse_succeeded
;

 total_old_parse_succeeded | total_new_parse_succeeded 
---------------------------+---------------------------
                    139522 |                    138903
```

New parser is able to parse slightly less feeds...

```sql
# SELECT
    (
        SELECT SUM(old_item_count)
        FROM scratch.feed_parser_comparison
        WHERE old_parse_succeeded
    ) AS total_old_item_count,
    (
        SELECT SUM(new_item_count)
        FROM scratch.feed_parser_comparison
        WHERE new_parse_succeeded
    ) AS total_new_item_count
;

 total_old_item_count | total_new_item_count 
----------------------+----------------------
              3796235 |              3840014
```

...but the new parser was able to find more items in feeds that it managed to parse.

For the feeds that both parsers managed to parse, feed's `<title>` length is about the same:

```sql
# SELECT
    AVG(old_feed_title_length) AS avg_old_feed_title_length,
    AVG(new_feed_title_length) AS avg_new_feed_title_length
FROM scratch.feed_parser_comparison
WHERE old_parse_succeeded AND new_parse_succeeded
;

 avg_old_feed_title_length | avg_new_feed_title_length 
---------------------------+---------------------------
       25.7582951367868777 |       25.5976101934330768
```

So is average length of `<title>`s of all items:

```sql
# SELECT
    AVG(old_total_items_title_length) AS avg_new_total_items_title_length,
    AVG(new_total_items_title_length) AS avg_new_total_items_title_length
FROM scratch.feed_parser_comparison
WHERE old_parse_succeeded AND new_parse_succeeded
;

 avg_old_total_items_title_length | avg_new_total_items_title_length 
----------------------------------+----------------------------------
            1554.5276384444636628 |            1542.3514010204964110
```

Item `<description>` lengths differ a bit more:

```sql
# SELECT
    SUM(old_total_items_description_length) / SUM(old_item_count) AS avg_old_total_items_description_length,
    SUM(new_total_items_description_length) / SUM(new_item_count) AS avg_new_total_items_description_length
FROM scratch.feed_parser_comparison
WHERE old_parse_succeeded AND new_parse_succeeded
;

 avg_old_total_items_description_length | avg_new_total_items_description_length 
----------------------------------------+----------------------------------------
                                   1004 |                                    934
```

I've looked into it, and it turns out that Python's `feedparser` does HTML (SGML) parsing for every `<description>` value and sanitizes non-whitelisted elements like `<script>`, also it gets rid of some whitespace, reorders element attributes and whatnot.

Average timestamps wary more wildly:

```sql
# SELECT
    AVG(old_average_item_timestamp)::bigint AS avg_old_average_item_timestamp,
    AVG(new_average_item_timestamp)::bigint AS avg_new_average_item_timestamp
FROM scratch.feed_parser_comparison
WHERE old_parse_succeeded AND new_parse_succeeded
;

 avg_old_average_item_timestamp | avg_new_average_item_timestamp 
--------------------------------+--------------------------------
                     1525763165 |                     1523778256
```

This skew originates from the fact that Perl's `XML::FeedPP` [returns publication date for RSS feeds and last edit date (if one is set) for Atom feeds](https://metacpan.org/pod/XML::FeedPP#$feed-%3EpubDate(-$date-);), while the Python feed parser (built on top of [`feedparser`](https://pythonhosted.org/feedparser/index.html)) returns an original publication date in all cases but resorts to last edit date if publication date is not set (which is rare if happens at all):

```sql
# SELECT
    (
        SELECT COUNT(*)
        FROM scratch.feed_parser_comparison
        WHERE old_parse_succeeded AND new_parse_succeeded
          AND old_average_item_timestamp = new_average_item_timestamp
    ) AS total_matching_timestamps,
    (
        SELECT COUNT(*)
        FROM scratch.feed_parser_comparison
        WHERE old_parse_succeeded AND new_parse_succeeded
          AND old_average_item_timestamp < new_average_item_timestamp
    ) AS total_new_timestamp_is_in_future,
    (
        SELECT COUNT(*)
        FROM scratch.feed_parser_comparison
        WHERE old_parse_succeeded AND new_parse_succeeded
          AND old_average_item_timestamp > new_average_item_timestamp
    ) AS total_new_timestamp_is_in_past
;

 total_matching_timestamps | total_new_timestamp_is_in_future | total_new_timestamp_is_in_past 
---------------------------+----------------------------------+--------------------------------
                    103058 |                              270 |                          16064
```

I think getting a publication date (as per the new behavior) is more fair because typically we assume that `stories.publish_date` represents a timestamp for when the story was first published (as reported by the website) and not for when we happened to fetch one of its edits. Anyway, a typical difference between publication date and modification date is only a few minutes.

---

So, overall it works fine! Good to merge? Here's the SQL dump of the stats if you'd like to have a look yourself:

[`scratch.feed_parser_comparison.sql.gz`](https://github.com/berkmancenter/mediacloud/files/4183851/scratch.feed_parser_comparison.sql.gz)

It's also available on our PostgreSQL.

hroberts:
this is great work both on the new feed parsing and on the validation.  please merge!

PR #655, opened 2020-01-10
`crawler-fetcher` rewritten to Python

pypt:
This is a rewrite of `crawler-fetcher` app to Python to make it easier to add podcast parsing additions later. The branch builds upon the previous [feed parsing rewrite to Python](https://github.com/berkmancenter/mediacloud/pull/653) which at the time of writing is still in testing.

Nothing particularly fancy or risky here, just a basic 1:1 rewrite.

The download "handler" has to fetch and store a download depending on its type. Different download types are split up into separate classes. I'm not particularly happy about how those classes are laid out (in Perl times, it was based on Moose "roles", now it's Python's "mix-in" pattern), I find them somewhat confusing, but I haven't thought of any better design.


hroberts:
this looks good to me.  obviously would be great to let this new code run for a few days somewhere to catch any errors.  otherwise good to merge.

pypt:
Deployed with a few hiccups.

Issue #650, opened 2020-01-08
implement support for podcast ingestion


rahulbot:
As per the plan developed in #515, we need to support podcast transcription (via Google). I'm creating this task to track implementation, as the other tracks planning.

As mentioned on the recent call, this needs to allow us to mark specific feeds / sources as a "podcast" source to queue if up for transcription. I was thinking this could be a need "feed.type" option, because that would make it easy to set in the UI:
![Update_Feed___The_Boston_Globe___Source_Manager___Media_Cloud](https://user-images.githubusercontent.com/673178/71995638-e6b57400-3208-11ea-88be-024dd41f1afe.jpg)

hroberts:
another option is to add a 'download and transcribe podcasts' boolean and
just use the existing 'syndicated' feed type.

pypt:
Deployed podcast transcription services, tried it with an initial podcast, seemed to work, so I've added the rest.

Here's the list:

https://docs.google.com/spreadsheets/d/1nUnhKGaazTgrUDhswCjz8KJ_KTlmSqvrF8VEkb2vQIs/edit#gid=0

All added podcast media sources are in `collection` tag set (`tag_sets_id = 5`) and are tagged with a `Podcasts` tag (`tags_id = 196654054`).

rahulbot:
I'm not sure they got added correctly. For instance, the [Glenn Beck podcast feed](https://sources.mediacloud.org/#/sources/1363086/feeds) show up as "syndicated" when I think it should be "podcast". Am I mistaken?

pypt:
Oh, and I was wondering why they weren't being fetched!

Updated to `podcast`, let's see if it works now.

rahulbot:
The new ['podcast' type feed on that source](https://sources.mediacloud.org/#/sources/1363086/feeds/1772783/edit) still hasn't fetched any stories. It looks like that is the Apple Podcasts URL, not the raw feed URL. Are you parsing out feed URLs from the Podcasts link automatically? 

To check this out I hacked somebody's script and made [a ruby script that fetches the feed url from the Apple Podcast URL](https://gist.github.com/rahulbot/bbc9750d8b1456bfd6de59231a2ac873). For that Glen Beck feed it spits out `https://feeds.megaphone.fm/BMDC3567910388` as the actual feed URL. Do we need to update the ones hosted at Apple Podcast to be the raw URLs or not?

pypt:
[...]

PR #656, opened 2020-01-21
Transcribe podcasts

pypt:
90% done but not quite finished podcast transcription PR.

Operates (will operate) more or less as follows:

1) When parsing a RSS / Atom feed, crawler's fetcher (`crawler-fetcher` app) will additionally write *all* found "enclosures" (`<enclosure />` in RSS, `<link rel="enclosure" />` in Atom) to `story_enclosures` table together with their URL, MIME type and size (if reported).
2) If podcast transcription is enabled for a particular feed (either `feeds.type = 'podcast'` or `feeds.is_podcast = 't'` -- still haven't decided which way to go), crawler's fetcher adds a `podcast-fetch-episode` RabbitMQ job for each newly found story in a feed;
3) **`podcast-fetch-episode` app:**
    1. Guesses the language the podcast episode will be most likely in based on `stories.title` + `stories.description` (both derived from RSS / Atom feed), converts the language code to BCP 47 identifier by additionally guessing the dialect the audio recording might be in (based on story's URL), e.g. we go for `en-AU` for stories originating from `.au` domains, `en-GB` for `.uk` ones, ..., and finally fall back on `en-US`;
    2. Chooses an enclosure from story's enclosures which is most likely a podcast episode (based on their reported MIME type and URL);
    3. Fetches said enclosure as a file and stores it locally; if the enclosure is more than 500 MB, the enclosure gets ignored;
    4. Transcodes a fetched enclosure if needed:
        * If enclosure's media file has a video stream, it gets discarded;
        * If enclosure's media file has multiple audio streams, all but one get discarded;
        * If enclosure's audio stream has more than one channel (e.g. it's stereo), all channels get mixed into one because Speech API can transcribe multi-channel audio files only if different speakers are recorded in each of the channels;
         * If enclosure's audio stream is not in one of the codecs that Speech API supports, or if any channel mixing has to be done, the whole audio stream gets transcoded to lossless FLAC codec in order to not lose any quality. I've read that ASR algorithms learn to do most of the guessing from sounds that the human ear can't even hear, and lossy codecs typically get rid of those sounds, so here we're trying to preserve whatever's left in that particular audio recording. FLAC is pretty space-efficient for a lossless codec - we can expect the final audio file to be around 2x the size of a stereo MP3 it was transcoded from;
    5. Uploads a transcoded audio file to Google Cloud Storage;
    6. Promotes a chosen story enclosure to an episode by writing a record to `podcast_episodes` table;
    7. Adds a `podcast-submit-operation` RabbitMQ job for the episode;
4) **`podcast-submit-operation` app:**
    1. Discards podcast episodes longer than two hours;
    2. Submits a Speech API job;
    3. Assumes that Speech API will take one minute of wallclock time to transcribe one minute of podcast episode audio;
    4. Adds a record to `podcast_episode_operations` table with Speech API's operation ID and a `fetch_results_at` timestamp set at `NOW() + podcast_duration`;
5) **`podcast-poll-due-operations` app:**
    1. Continuously polls `podcast_episode_transcript_fetches` table looking for transcription operations which should be finished already (based on their `add_to_queue_at` value);
    2. Adds likely-to-be-done transcription operations to `podcast-fetch-transcript` RabbitMQ queue as jobs;
6) **`podcast-fetch-transcript` app:**
    1. Tries to fetch the results of a Speech API operation in order to retrieve podcast episode's transcription;
    2. Updates `podcast_episode_transcript_fetches.fetched_at` to current timestamp so that we don't try to fetch it again.
    3. If the operation is finished already:
        * If the operation succeeded:
            * Stores the transcript in raw downloads S3 bucket;
            * Adds an `extract-and-vector` job for it to fetch the raw download, "extract" it (there's nothing to extract really) and do the rest of the stuff;
            * (We could write the transcribed text directly to `download_texts` but that would break many assumptions about what's stored in `downloads` and S3, so we might, for example, mess up a mass reextraction in the future.)
        * If the operation failed:
            * Stores the error message in `podcast_episode_transcript_fetches.error_message` column.
    4. If the operation is not finished yet:
        * Sets the operation's `fetch_results_at` to `NOW() + '10 minutes'` for the `podcast-poll-due-operations` to re-add the `podcast-fetch-transcript` job in another 10 minutes.
    5. Updates `podcast_episode_transcript_fetches.result` with `in_progress` / `success` / `error` transcript fetch attempt result.

It might look like a lot of new apps for something that could be one by a single script, but this setup is more dynamic, e.g. at some point we might decide that we'd like to download the episodes only and skip the costly transcribing for now, or we might want to allot CPU / RAM resources differently and have different number of workers for every step of the chain, or we could have only one `podcast-submit-operation` worker to limit API calls to the Speech API, etc.

As for testing, there are a couple of integration tests under `podcast-fetch-transcript` which submit a story enclosure on one end, wait a bit for all of the other apps to do their work and see what came out at the other end, and also a bunch of unit tests for every sub-step in the process, e.g. we test out transcoding with various media file inputs.

There's only a single document for now describing how to create a Google Cloud service account for podcast transcription use here: https://github.com/berkmancenter/mediacloud/blob/transcribe_podcasts/doc/podcasts_gc_auth.markdown. Code is somewhat documented too.

As per Facebook client Python rewrite example (https://github.com/berkmancenter/mediacloud/pull/640), all of the apps make a distinction between "soft" errors and "hard" errors, and `exit(1)` on the latter.

References #650.

Depends on:

* https://github.com/berkmancenter/mediacloud/pull/653 (still needs to be tested with some live data and then merged)
* https://github.com/berkmancenter/mediacloud/pull/655 (needs merging)


pypt:
Ready for deployment (once we do validation for https://github.com/berkmancenter/mediacloud/pull/653). Some changes:

* Podcast feeds get their own `feeds.type = 'podcast'` (ping @rahulbot)

* Supported feed URL types are:
    * A simple RSS / Atom feed URL pointing to a podcast, e.g. https://www.residentadvisor.net/xml/podcast.xml
    * iTunes Podcasts URL, e.g. https://podcasts.apple.com/lt/podcast/ra-podcast/id129673441
    * Google Podcasts URL, e.g. https://podcasts.google.com/?feed=aHR0cHM6Ly93d3cucmVzaWRlbnRhZHZpc29yLm5ldC94bWwvcG9kY2FzdC54bWw

PR #662, opened 2020-02-07
Import UTF-8 stories into Solr correctly

pypt:
Hey Hal,

Another year, another "we've been importing garbled UTF-8 content into Solr, and no one noticed for quite some time" bug!

Research team reported that Hindi language queries didn't seem to work here:

https://github.com/mitmedialab/MediaCloud-Web-Tools/issues/1731#issuecomment-583146920

I looked into it, and turns out that a lot of Hindi and other UTF-8 text was garbled (double-encoded) in Solr.

Here's the proposed fix: in `set_content()`, try to identify whether Perl already thinks that scalars are in UTF-8, and encode them accordingly. Added a few UTF-8 tests as well, plus verified that stuff gets imported into Solr with the right encoding manually.

After deploying this, we'll need to reimport a bunch of stories, but it's hard to measure the impact of the bug - in the linked issue, it seems that some stories stopped coming in at around October or November, so maybe this happened due to me containerizing things? Or could those "other" stories be something that we've imported into Solr through some other means? I'm not sure.

### DIFF
--- a/apps/common/src/perl/MediaWords/Solr.pm
+++ b/apps/common/src/perl/MediaWords/Solr.pm
@@ -208,7 +208,7 @@ sub query_encoded_json($$;$)
 
     $params->{ fq } = [ map { _insert_collection_media_ids( $db, $_ ) } @{ $params->{ fq } } ];
 
-    my $response_content = MediaWords::Solr::Request::solr_request( 'select', {}, $params, 'application/x-www-form-urlencoded' );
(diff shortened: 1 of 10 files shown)


pypt:
Just going to merge, reimport everything since September and see what happens.

pypt:
Deployed, started readding 134m stories since September into `solr_import_stories` in chunks.

pypt:
Reimport finished, Hindi queries working again.

PR #730, opened 2020-09-29
In attach_story_data_to_stories(), don't overwrite existing fields

pypt:
Hey Hal!

I know you're superbusy with other stuff, but could you have a quick look?

In https://github.com/mediacloud/backend/issues/729 (and possibly https://github.com/mediacloud/backend/issues/725), it looks like `MediaWords::DBI::Stories::attach_story_data_to_stories()` used to reset `$list_field` key to an empty arrayref on every call, thus making this subroutine not work when it was called multiple times, e.g. when data was being "attached" in chunks:

https://github.com/mediacloud/backend/blob/12a0c0e896bb6979469248d43f4a7f1ec23c3d8e/apps/webapp-api/src/perl/MediaWords/Controller/Api/V2/StoriesBase.pm#L279-L302

I've now made the subroutine *merge* the input `$story_data` with whatever might already be in the parameter `$stories` under `$list_field` (if anything). There's also a test that verifies that the subroutine behaves as intended.

Does it look good to you?

### DIFF
--- a/apps/common/src/perl/MediaWords/DBI/Stories.pm
+++ b/apps/common/src/perl/MediaWords/DBI/Stories.pm
@@ -35,7 +35,21 @@ sub attach_story_data_to_stories
 {
     my ( $stories, $story_data, $list_field ) = @_;
 
-    map { $_->{ $list_field } = [] } @{ $stories } if ( $list_field );
+    if ( $list_field ) {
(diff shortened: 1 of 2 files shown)


hroberts:
looks good to me. nice catch.

PR #745, opened 2020-12-10
merge cliff apps + nytlabels apps

jtotoole:
For https://github.com/mediacloud/backend/issues/743. Questions/comments:

- The tests for `cliff-fetch-annotation-and-tag` + `nytlabels-fetch-annotation-and-tag` are working. I don't have tests for the new `tag-from-annotation` module in `mediawords`. I was assuming that's kosher because, similar to `mediawords/annotator`, it's just abstract classes, and the child classes do get tested. Am I correct about that?
- [`.../perl/MediaWords/Annotator/Store.pm`](https://github.com/mediacloud/backend/blob/master/apps/common/src/perl/MediaWords/Annotator/Store.pm#L19) is the only place in the project I see referencing the Python [`.../mediawords/annotator`](https://github.com/mediacloud/backend/tree/master/apps/common/src/python/mediawords/annotator) directory subsequent to these changes. Can `Store.pm` be safely deleted (and with it `.../python/mediawords/annotator/*`), or does that file need to be preserved/changed?
- I had to change a couple package versions in [this file](https://github.com/mediacloud/backend/blob/master/apps/nytlabels-annotator/src/crappy-predict-news-labels/requirements.txt) to be compatible with `tensorflow==1.15.4` in order to get `nytlabels-annotator` to build, but you may have a better approach in mind to deal with that.
- I was thinking of doing the migrations in a separate PR if that's okay with you.

No rush at all on this—thx!


apps/common/src/python/mediawords/tag_from_annotation/fetch_and_tag.py
```diff
@@ -0,0 +1,403 @@
+import abc
```
pypt, line 1: Heh, it turns out that there are API endpoints which return those raw CLIFF / NYTLabels annotations. So that was the reason why we were storing raw responses!

Good news is that I'm 99,998% positive that no one uses those raw responses for anything, and even if they were using those, it would be easier just to ask CLIFF / NYTLabels annotators directly for them to generate a new raw response.

So could you do the following:

1. Confirm with Dennis that those API calls (https://github.com/mediacloud/backend/blob/master/doc/api_2_0_spec/admin_api_2_0_spec.md#apiv2storiescliff and https://github.com/mediacloud/backend/blob/master/doc/api_2_0_spec/admin_api_2_0_spec.md#apiv2storiesnytlabels) are not used anywhere;
    * If they are, then `break`, ping me and I'll rewrite Perl API calls for them to get those raw annotations from the services which generate them.
    * If they are not, then `continue`
2. Delete everything that references `mediawords.annotator` Python module, including the module itself;
3. Delete everything that references `MediaWords::Annotator` Perl package, including the package itself. I realize Perl is not your speciality, but deleting a Perl package should be rather straightforward - get rid of https://github.com/mediacloud/backend/tree/master/apps/common/src/perl/MediaWords/Annotator and remove everything in code that references `MediaWords::Annotator` (including whole `sub cliff` and `sub nytlabels` subroutines that "serve" the API call). Let me know if you need my help with that.
4. Update the API spec (https://github.com/mediacloud/backend/blob/master/doc/api_2_0_spec/admin_api_2_0_spec.md) for it to no longer include the references to those two removed API calls.
    * You can use `./doc/update_toc.sh` to update Table of Contents of the API spec after adding / removing some sections.

pypt (changes requested):
Thanks James! Just a few small fixes plus some more extra deleting for you to do.

Tests look okay to me. Yep, the Perl stuff has to go (refer to the comment in the review). TensorFlow thing should work too. Regarding dropping the tables, please do whatever is more convenient to you.

apps/cliff-fetch-annotation-and-tag/.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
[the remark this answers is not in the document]
jtotoole: I think this is just a relic of doing a `cp -r` from one of the existing cliff apps, so I'm happy to just get rid of it

apps/common/src/python/mediawords/tag_from_annotation/fetch_and_tag.py
jtotoole: Regarding 1:

```
Dennis  < 1 minute ago
I can confirm that both are indeed used
```

😞 

No rush on the Perling, I can still get this branch ready in the meantime

apps/cliff-fetch-annotation-and-tag/bin/cliff_tags_from_annotation_worker.py
```diff
@@ -1,26 +1,26 @@
 #!/usr/bin/env python3
```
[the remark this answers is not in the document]
jtotoole: The containers start and seem to execute the worker files correctly when I do `docker run -it dockermediacloud/cliff-fetch-annotation-and-tag`; however, I do get a permissions error with `docker-compose -f apps/cliff-fetch-annotation-and-tag/docker-compose.tests.yml up`:

```
Starting cliff-fetch-annotation-and-tag_postgresql-server_1 ... done
Starting cliff-fetch-annotation-and-tag_postgresql-pgbouncer_1 ... done
Recreating cliff-fetch-annotation-and-tag_cliff-fetch-annotation-and-tag_1 ... done
Attaching to cliff-fetch-annotation-and-tag_postgresql-server_1, cliff-fetch-annotation-and-tag_postgresql-pgbouncer_1, cliff-fetch-annotation-and-tag_cliff-fetch-annotation-and-tag_1
...
cliff-fetch-annotation-and-tag_1  | [FATAL tini (8)] exec cliff_tags_from_annotation_worker.py failed: Permission denied
```

Any idea what might be going on there? I added the `chmod` command [here](https://github.com/mediacloud/backend/pull/745/files#diff-d25bd683c929f54f906981ae23c80473ece4041950f449c05e17e56cc14f83a3R16).

Side note: The existing cliff/nytlabels apps [don't appear to use `chmod`](https://github.com/mediacloud/backend/blob/master/apps/cliff-update-story-tags/Dockerfile#L13-L17) in their Dockerfiles. Is there a difference that necessitates it in this case?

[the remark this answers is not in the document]
pypt: It's not the `RUN chmod` command that you need to add, you need to `chmod +x` the files in the Git repo and then commit it. Git will then preserve the executable bit when checking out, and you won't have to do it in the `Dockerfile`.

apps/common/src/python/mediawords/tag_from_annotation/fetch_and_tag.py
pypt: Darn it :( Thanks for asking Dennis. Okay, then I'll replace those two API endpoints with something that calls the `*-annotator` service and we could take it from there.

apps/cliff-fetch-annotation-and-tag/bin/cliff_tags_from_annotation_worker.py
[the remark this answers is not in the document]
jtotoole: ahhhhhh got it! `docker compose` commands successfully starting the workers now

jtotoole:
alrighty, i think i've got all the review feedback addressed. i'll put the merge on hold pending your API change—much appreciated!

apps/webapp-api/src/perl/MediaWords/Controller/Api/V2/Stories.pm
```diff
@@ -12,7 +12,7 @@ use HTTP::Status qw(:constants);
 use Readonly;
 use Encode;
 
-use MediaWords::Annotator::Store;
+# use MediaWords::Annotator::Store;
```
jtotoole, line 15: are the changes required here as simple as deleting what i've got commented out at present?

pypt: Yeah, just delete everything `MediaWords::Annotator` related and the package itself please.

pypt:
Now that the API calls have been made to call the underlying services directly, could you resolve these merge conflicts? I see some of the conflicts in Perl code as well so let me know if you need my help with that.

jtotoole:
okay all set here i think, though i'd be happy to have you give the changes another quick look

doc/api_2_0_spec/admin_api_2_0_spec.md
```diff
@@ -13,12 +13,6 @@ Table of Contents
       * [api/v2/stories/update (PUT)](#apiv2storiesupdate-put)
          * [Input Description](#input-description)
          * [Example](#example-1)
-      * [api/v2/stories/cliff](#apiv2storiescliff)
```
pypt, line 16: Could you restore those two API calls in the docs (both the T&C and the actual entry below)? Those API calls (`api/v2/stories/cliff` and `api/v2/stories/nytlabels`) still exist, the only change is that now they hit the underlying `cliff-annotator` and `nytlabels-annotator` instead of reading the annotations from disk.

doc/api_2_0_spec/admin_api_2_0_spec.md
```diff
@@ -213,112 +207,6 @@ Output:
 { "success": 1 }
 ```
 
-## api/v2/stories/cliff
-
```
pypt, line 217: These too!

I don't have enough brain capacity to understand Git's manpages so I typically use one of the Git's GUI tools for that. When I was using a Mac, my favorite was [Sourcetree](https://www.sourcetreeapp.com/) (free!) but then my beloved Mac almost caught fire so I resorted to [Sublime Merge](https://www.sublimemerge.com/) (paid, but doesn't beg for $ too often).

Fastest way to recover these accidentally removed blocks with a GUI tool would be to find out the commit that got them removed, search / scroll down to that commit on a GUI tool, select the lines that you want to restore, click on "Revert" or something like that, and commit the change once you're happy with it. You can do it [using the terminal too](https://stackoverflow.com/a/5669428/200603).

pypt (changes requested):
Superb, except for docs for those two API calls that got removed. Could you restore them real quick?

jtotoole:
all set i think!

pypt:
Thank you James!

PR #747, opened 2020-12-15
Add topic mine support for the Pushshift verified twitter archive

epenn:
This adds support for pulling data from the Pushshift verified twitter archive. A couple things of note:

- Implemented using Elasticsearch's scroll API for paging support.
- This makes a second round trip after getting the results to fill in `retweeted_status` and `quoted_status` fields (when needed) since Pushshift optimizes for space by removing the payloads from quote tweets and retweets.

### DIFF
--- a/apps/topics-mine/src/python/topics_mine/posts/pushshift_twitter.py
+++ b/apps/topics-mine/src/python/topics_mine/posts/pushshift_twitter.py
@@ -1,0 +1,291 @@
+"""Fetch verified twitter posts from Pushshift."""
+
+import functools
+from datetime import datetime
+
+import dateutil.parser
+import requests_mock
+
+from mediawords.util.parse_json import decode_json, encode_json
+from mediawords.util.web.user_agent import UserAgent
+from mediawords.util.web.user_agent.request.request import Request
+
+from topics_base.posts import get_mock_data
+from topics_base.twitter_url import get_tweet_urls
+from topics_mine.posts import AbstractPostFetcher
+
+from mediawords.util.log import create_logger
+log = create_logger(__name__)
+
+PS_TWITTER_PAGE_SIZE = 10000
+PS_TWITTER_SCROLL_TIMEOUT = '1m'
+PS_TWITTER_SCROLL_URL = 'https://twitter-es.pushshift.io/_search/scroll'
+PS_TWITTER_URL = 'https://twitter-es.pushshift.io/twitter_verified/_search?scroll=%s' % PS_TWITTER_SCROLL_TIMEOUT
+
+
+class McPostsPushshiftTwitterDataException(Exception):
+    """Error while interrogating the response from the Pushshift API."""
+    pass
+
+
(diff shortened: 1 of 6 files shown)


pypt:
@epenn would you be able to rebase this off the top of the current `master` branch?

apps/topics-mine/src/python/topics_mine/posts/pushshift_twitter.py
```diff
@@ -0,0 +1,291 @@
...
+        """Fetch quote tweets and retweets referenced by tweets in the list.
+
+        Since Pushshift doesn't store the referenced tweet dict for quote tweets or retweets, only
+        their IDs, a second roundtrip is necessary to complete records with a referenced tweet.
+        
+        For example, on a quote tweet the quoted_status field isn't present, only quoted_status_id.
+        Similarly, on a retweet the retweeted_status field isn't present, only retweeted_status_id.
+
+        Returns: None, referenced tweets are stored in tweet['quoted_status'] or
+        tweet['retweeted_status'] as appropriate.
```
pypt, line 181: So this modifies the parameter `tweets` in-place, right?

Could you make it return the modified tweets instead? Now it's a bit hard to figure out which specific method does add those `quoted_status` / `retweeted_status` to the `tweets`. We have a plenty of those C-style "pass a reference" methods that change up their arguments in subtle ways all over our codebase, and I can't say I like them too much :)

apps/topics-mine/src/python/topics_mine/posts/pushshift_twitter.py
```diff
@@ -0,0 +1,291 @@
...
+
+        posts = []
+        for tweet in tweets:
+            log.debug('tweet: %d' % tweet['id'])
+            publish_date = dateutil.parser.parse(tweet['created_at']).isoformat()
+            posts.append({
+                'post_id': tweet['id_str'],
+                'data': tweet,
+                'content': tweet['text'],
+                'publish_date': publish_date,
```
pypt, line 253: Could you verify that `publish_date` gets converted to `America/New_York` timezone at some point, in the caller perhaps? Unfortunately `topic_posts.publish_date` is a `TIMESTAMP [WITHOUT A TIME ZONE]` in PostgreSQL, so we have to normalize all the dates that come in to EST / EEST (or maybe it's UTC for `topic_posts` specifically? Either way, it has to point to the same moment of time somehow), and I don't know whether it's the caller that does that, or are you supposed to do it in the post ingest module such as this one.

pypt (changes requested):
Thank you for comprehensible code comments Eric!

For the reference, could you also ELI5 what this PR is all about, i.e. what's "verified Twitter" archive and what is it that we're going to be doing here? I genuinely don't know :)

pypt:
Oh, and I have merged in `master`, hope that's okay.

pypt:
Could you also have a look at the failing tests too?

pypt:
Thanks Rahul!

PR #753, opened 2020-12-29
Python mine

hroberts:
This is the migration of the topics-mine and topics-mine-public workers from perl to python.  This is almost entirely just a line by line migration, with only a few small fixes / refactors as I went.  The big change is adding unit tests for all non-trivial code in mine.py.  I also migrated the existing big integration test that was in test_tm_mine.t to test_mine.py.

I think test_mine.py passed at the first try once I got the test itself working, which I take as a good sign that the unit test coverage is good.  I'm still nervous about plugging this directly into public topics, so I plan on turning off the worker pools and running a few production topics from the command line before turning the pools back on.


jtotoole:
Still knowing very little about how topics work, I don't feel qualified to opine on this at great length, but it looks well-organized and well-tested. Nice work! Looking forward to hearing how the test topics from the CLI turn out.

jtotoole (approved):

PR #758, opened 2021-01-22
Update Ubuntu to 20.04, Python to 3.8; nytlabels-annotator ONNX models, resource usage improvements

pypt:
I've updated the containers to inherit from `ubuntu:focal` (Ubuntu 20.04) instead of `ubuntu:xenial` (Ubuntu 16.04) because:

* 16.04 won't be supported anymore after April;
* host systems run 20.04 so it's awkward to run 20.04 on host and 16.04 on userland;
* ELK stack for aggegating logs that's implemented in another branch will want to ingest journald logs from the host system, and the host system runs journald which is incompatible with 16.04's journald log format.

Ubuntu upgrade comes with a Python 3.7 -> 3.8 upgrade as 20.04 ships with that.

Most of the huffing and puffing had to be done to the `nytlabels-annotator`: the app uses a bunch of ML models that were trained in prehistoric times with TensorFlow 0.x and by pure chance still managed to load with TF 1.x, however, they don't work with TF 2.x anymore, and TF 1.x does not (nor it ever will) support Python 3.8.

At first I tried to retrain the model with TF 2.x - I made [Jasmin's training code](https://github.com/mediacloud/nytlabels-annotator-train) work again by rewriting it to use SQLite3, and then fed it the [New York Times Annotated Corpus](https://catalog.ldc.upenn.edu/LDC2008T19), but no matter what I tried, my retrained model achieved only 40% or so accuracy, and the resulting labels didn't look nothing like the ones from the current model; my guess is that either the training data was custom-tailored in some way, or I'm just not that good at ML stuff (which is the case).

Then, I tried to export the TF 1.x model to [ONNX format](https://onnx.ai/), but it turns out that the tooling that converts stuff to ONNX didn't really support TF 1.x. Lastly, I reimplemented the model in TF 2.x, exported the weights from the old TF 1.x model, shuffled some dimensions in them using the timeless

i-dont-know-what-im-doing

method, loaded the weights into my guessworked TF 2.x model, and exported it to ONNX. The effort resulted with models that can be loaded anywhere, use a very lean runtime (no more 20+ Python modules to install), don't depend on a specific version of said runtime, and predict identical labels (with very similar scores).

I've also exported word - vector pairs from the [Google News word2vec corpus](https://github.com/mmihaltz/word2vec-GoogleNews-vectors) that we use in `nytlabels-annotator` into a [`shelve` database](https://docs.python.org/3/library/shelve.html), and trimmed it a bit by removing multi-word pairs from it (the model's word tokenizer splits the sentence only into single words so those word pairs would not have ever been matched). The resulting Docker image is now smaller (5.79 GB -> 1.89 GB), uses less RAM (8 GB -> 1 GB), loads pretty much instantly (as it doesn't have to load the whole word2vec corpus into RAM), and is insignificantly slower (word2vec mappings now have to be looked up from disk). Ping @rahulbot, maybe you'd like to copy these `nytlabels-annotator` changes to your own repo.

Some more notes:

* still working on fixing the tests;
* RabbitMQ's version got bumped from 3.7.7 to 3.8.9 which is unlikely to be a big deal, but note to self: make backup of RabbitMQ's data partition before deploying;
* added `/dl_to_stdout.sh` script which is just a shorthand for `curl --fail --location --retry 3 --retry-delay 5` (useful for downloading huge files from potentially wonky sources).
* when using home-made ML models, we should make sure to preserve both the code that created the model and the data (or at least instructions on re-acquiring the data) that was used to train the model; it's a whole week's ordeal to make such a black box model work again after a few years pass;
* I think it's time to let someone make us a brand new multi-label news article classifier; our current `nytlabels-annotator` was made in 2017 using articles from 1987 to 2007 so it's probably not making good guesses about articles on TikTok or Yeezy; plus, since 2017 new ML stuff came up such as [BERT](https://github.com/google-research/bert), new datasets ([Yahoo News](https://yahooresearch.tumblr.com/post/164789421126/introducing-the-yahoo-news-ranked-multi-label), probably others), etc.; maybe it could be a task for the GSoC student?

Diffs between TF 1.x and ONNX labels / scores for a sample text:

```diff
--- all_descriptors-old.txt	2021-01-21 20:42:08.584501960 +0200
+++ all_descriptors-new.txt	2021-01-22 18:41:12.268918517 +0200
@@ -1,12 +1,12 @@
   * Label: politics and government, score: 0.109894
-  * Label: recordings (video), score: 0.091937
+  * Label: recordings (video), score: 0.091936
   * Label: television, score: 0.087924
   * Label: freedom and human rights, score: 0.087888
   * Label: demonstrations and riots, score: 0.069053
   * Label: freedom of speech and expression, score: 0.065840
   * Label: computers and the internet, score: 0.053404
   * Label: no index terms from nytimes, score: 0.051645
-  * Label: games, score: 0.050916
+  * Label: games, score: 0.050917
   * Label: motion pictures, score: 0.038354
   * Label: speech, score: 0.034407
   * Label: recordings (audio), score: 0.028925
```

```diff
--- descriptors3000-old.txt	2021-01-21 20:41:30.408619699 +0200
+++ descriptors3000-new.txt	2021-01-22 18:41:23.360871517 +0200
@@ -1,4 +1,4 @@
-  * Label: censorship, score: 0.420402
+  * Label: censorship, score: 0.420403
   * Label: television, score: 0.289587
   * Label: demonstrations and riots, score: 0.187069
   * Label: computers and the internet, score: 0.070932
@@ -7,7 +7,7 @@
   * Label: politics and government, score: 0.045939
   * Label: freedom of speech and expression, score: 0.035946
   * Label: shows (exhibits), score: 0.034075
-  * Label: recordings and downloads (video), score: 0.032530
+  * Label: foreclosures, score: 0.032530
   * Label: advertising and marketing, score: 0.030223
   * Label: children and youth, score: 0.027249
   * Label: motion pictures, score: 0.022378
```

```diff
--- descriptors600-old.txt	2021-01-21 20:41:04.216700298 +0200
+++ descriptors600-new.txt	2021-01-22 18:41:33.644827983 +0200
@@ -13,7 +13,7 @@
   * Label: entertainment and amusements, score: 0.042958
   * Label: games, score: 0.042757
   * Label: advertising, score: 0.041260
-  * Label: media, score: 0.038266
+  * Label: media, score: 0.038267
   * Label: economic conditions and trends, score: 0.030450
   * Label: no index terms from nytimes, score: 0.027031
   * Label: athletics and sports, score: 0.026866
```

```diff
--- descriptors_with_taxonomies-old.txt	2021-01-21 20:42:32.048429440 +0200
+++ descriptors_with_taxonomies-new.txt	2021-01-22 18:41:43.100787982 +0200
@@ -1,5 +1,5 @@
   * Label: top/news, score: 0.849702
-  * Label: top/features/travel/guides/destinations/asia/china, score: 0.625038
+  * Label: top/features/travel/guides/destinations/asia/china, score: 0.625039
   * Label: motion pictures, score: 0.607049
   * Label: top/features/movies, score: 0.531915
   * Label: top/news/world/countries and territories/china, score: 0.388632
@@ -13,7 +13,7 @@
   * Label: no index terms from nytimes, score: 0.078892
   * Label: top/news/business, score: 0.075088
   * Label: games, score: 0.064192
-  * Label: top/news/business/media and advertising, score: 0.049234
+  * Label: top/news/business/media and advertising, score: 0.049235
   * Label: recordings (video), score: 0.046624
   * Label: reviews, score: 0.045932
   * Label: top/news/world, score: 0.039019
```

```diff
--- just_taxonomies-old.txt	2021-01-21 20:42:48.616378164 +0200
+++ just_taxonomies-new.txt	2021-01-22 18:41:50.344757353 +0200
@@ -1,7 +1,7 @@
   * Label: Top/News, score: 0.687006
   * Label: Top/Features/Travel/Guides/Destinations/Asia/China, score: 0.683852
   * Label: Top/Classifieds/Job Market/Job Categories/Media, Entertainment and Publishing, score: 0.296769
-  * Label: Top/Features/Travel/Guides/Destinations/Asia, score: 0.253783
+  * Label: Top/Features/Travel/Guides/Destinations/Asia, score: 0.253784
   * Label: Top/News/World/Countries and Territories/China, score: 0.249016
   * Label: Top/News/Business, score: 0.196076
   * Label: Top/News/World/Asia Pacific, score: 0.144047
```


pypt:
> I just started my teaching for the semester, so I wanted to note that I probably won't be able to get to this for at least another week or two.

No problem, we still use my "crappy predict news labels" reimplementation of the predictor on the backend, so this is just a heads-up.

jtotoole (approved):

pypt:
* Updated some Python dependencies
* It seems that we're stuck with Celery 4.x for now because Celery 5.x changed up a bunch of things and so our [`mediawords.job.JobBroker` wrapper around Celery](https://github.com/mediacloud/backend/blob/7e8529c5765ecc86e22d456750f0d78b441901ec/apps/common/src/python/mediawords/job/__init__.py#L374) no longer works
* New NLTK versions have a slightly different opinion on how to tokenize Norwegian and Hungarian :)
* [`test_remote_platforms.py`](https://github.com/mediacloud/backend/blob/7e8529c5765ecc86e22d456750f0d78b441901ec/apps/topics-mine/tests/python/test_remote_platforms.py#L17) was fetching 16k tweets from Brandwatch and thus making 1600 Twitter API calls; now it doesn't do that much
* Removed "googler" platform as it didn't seem to work and was commented out everywhere

Issue #754, opened 2021-01-07
Replace home-grown PostgreSQL database migrations with something modern


pypt:
Hey James!

So, are you up for fixing up some past mistakes of mine?

As you might have noticed (and as is described in the [docs](https://github.com/mediacloud/backend/blob/master/doc/schema_migrations.markdown)), we use a self-made database migrations system that I came up with. The essence of it is that to come up with a migration, one has to:

1. Update `apps/postgresql-server/schema/mediawords.sql` which serves as our reference schema file.
2. Add a new migration under `apps/postgresql-server/schema/migrations/mediawords-XXX0-XXX1.sql` that will get imported when the production database actually get migrated.

This process kinda works, but there are a few issues with it:

* Migrations don't get automatically tested at any point. When `postgresql-server` gets built into a Docker image, only the reference `schema/mediawords.sql` gets imported which is not something that we do at all when deploying. Instead, migrations get applied on `postgresql-server` container start, i.e. the `bin/postgresql_server.sh` (via `bin/apply_migrations.sh`) starts the server in an unpublished port (IIRC `1234`), tests whether the schema of the database is up to date, applies the migrations if needed, stops the server and then restarts in on a "proper" port. If a migration fails, then the container never really starts (which is okay because we don't want a database instance with an out-of-date schema but also not okay because we should somehow learn that the migration is broken before we even try deploying it).
* Duplicate work. To add a table, one has to edit both `schema/mediawords.sql` and a migration file in `schema/migrations/`.

So, a new migration system is due. Something like this would be a tremendous improvement IMHO:

1. We get rid of the `mediawords.sql` file altogether and manage the schema only via migrations.
2. At build time, we import all the migrations sequentially to end up with a live schema.
3. At runtime, we maybe follow a similar approach to what's currently being done - start a "private" instance of PostgreSQL, run a migration tool against it to get it up-to-date, and then restart it to make it public.

Or is there a better way to do migrations these days?

Vague to-do:

1. Come up with a few migration tools that you like, describe them in this issue and pick one that you think we should go with (I mentioned [Flyway](https://flywaydb.org/) during our meeting but it doesn't have to be Flyway at all, just something that makes the most sense to you).
2. (Attempt to) rewrite current migrations to said tool.
[...]

PR #759, opened 2021-01-24
[RFR] use pgmigrate for schema migrations

jtotoole:
Progress here: pgmigrate is successfully creating the db and running the migrations on container start. Some questions (apologies in advance that these reflect a shaky grasp of the existing system and are therefore likely dumb):

- `pgmigrate` does the work both of initializing the schema and applying migrations, so I'm thinking that `initialize_schema.sh`, `apply_migrations.sh`, and `generate_empty_sql_migration.sh` can perhaps be consolidated into one file. The challenge, then, is whether there's a way of running a check, like what's done [here](https://github.com/mediacloud/backend/blob/88618c545efbb9b9744ea89d18fe413d7a3a5bb4/apps/postgresql-server/bin/apply_migrations.sh#L64-L89), to determine whether new migrations are necessary on container start. Do you have thoughts on that? Maybe running all the migrations every time is unavoidable? One thing that `pgmigrate` can do is point to a specific migration number as opposed to applying them all at once (e.g `pgmigrate -t 3 migrate` to run up to migration 3), and it creates a `schema_version` table in the DB. So, maybe there's a way of scanning all the files in the `migrations` folder, identifying the filename with the highest number, and comparing that to the highest number in the `schema_versions` table?
- It seems important to [validate whether the ZFS backup snapshot works,](https://github.com/mediacloud/backend/blob/master/apps/postgresql-server/bin/postgresql_server.sh#L17-L19), however I don't exactly grok what's happening in that code block. How do you think that check might be incorporated into the new system?
- You noted [in the issue](https://github.com/mediacloud/backend/issues/754):

> It would be nice to have (retain) to have some way of applying any given migrations manually before deploying the rebuilt postgresql-server service.... If the migration tool would be able to print out SQL that would be run on the live database instead of the tool insisting that it has to run the SQL itself, that would be pretty great!
- ^pgmigrate has a `--dryrun` option, which rolls back rather than committing, but it doesn't seem to actually log the SQL anywhere when that flag is set, so I'm not sure of the best way to output the pending SQL code without running it. I think, though, that all it's going to do each time is run the files in `/migrations` sequentially?  
- I'm considering the best way to run `pg_dump` to get the reference schema file. Based on the permissions errors I've been getting when attempting to run as `root` and `postgres`, as well as a read [of the Postgres docs](https://www.postgresql.org/docs/11/backup-dump.html), it seems like I need to execute the command as the `mediacloud` superuser. When I've tried that (specifically, executing `pg_dump --dbname=mediacloud --username=mediacloud` in `initialize_db.sh`), I get the error: `pg_dump: [archiver (db)] connection to database "mediacloud" failed: FATAL:  Peer authentication failed for user "mediacloud"`. Any thoughts on how to solve this one?

Perhaps it's easiest to talk through this via Google Meet—lmk!


pypt:
> I'm thinking that initialize_schema.sh, apply_migrations.sh, and generate_empty_sql_migration.sh can perhaps be consolidated into one file

`generate_empty_sql_migration.sh` is a script that creates a bunch of boilerplate for a new migration, so given that there's not going to be any boilerplate to wrap the actual migration in, I don't think it's necessary to have that script anymroe.

`initialize_schema.sh` preloads `/var/lib/postgresql/` with a empty schema. We employ a somewhat neat Docker trick here: `/var/lib/postgresql/` [is defined as a volume in `Dockerfile`](https://github.com/mediacloud/backend/blob/88618c545efbb9b9744ea89d18fe413d7a3a5bb4/apps/postgresql-server/Dockerfile#L66), and if you write something to a directory (`/var/lib/postgresql/`) and later declare it as a volume, on a container start Docker will then:

1) If no volumes from the host system are attached to the container, use `/var/lib/postgresql/` as any other directory;
2) If a newly created volume gets attached to the container, preload that new volume with the contents of `/var/lib/postgresql` from the container image;
3) If a previously created volume gets attached to the container, don't preload anything and just start the container with the volume mounted.

That way we can use the same `postgresql-server` image for both testing and production:

1) When running a test (or anything else that gets started via `docker-compose.tests.yml`), no volumes get mounted to `/var/lib/postgresql/` in `postgresql-server`, so the container starts with an empty schema loaded to it via `initialize_schema.sh`;
2) When starting the very same Docker image on production, a data volume gets mounted from the host to the container so PostgreSQL gets started with the production data instead of an empty schema that got previously built into the container image.

Maybe there's a better way to go about all of this, but to retain the current functionality / behaviour, you'd have to make sure of the following:

* On Docker build, the image ends up with the empty, up-to-date schema preloaded under `/var/lib/postgresql`. To do that, you'd probably want to start PostgreSQL in the background, apply all the migrations in the sequence to it, and stop it (as is currently done in `initialize_schema.sh`).
* On container start, whenever container comes around to opening up a public port (i.e. lets users connect), it should expose an up-to-date schema to the users. To achieve that, you might start PostgreSQL on a non-public port (to avoid letting users connect to an outdated database), run the migration tool against that instance (which would then in turn figure out at which "version" the currently live schema is at, and apply the missing migrations to make the schema up-to-date), stop the instance, and restart it on a public port for the rest of the services to use.

As for the `schema_version` table, I think the tool is supposed to do all of the version comparing and such itself?

> It seems important to validate whether the ZFS backup snapshot works,, however I don't exactly grok what's happening in that code block. How do you think that check might be incorporated into the new system?

`MC_POSTGRESQL_SKIP_MIGRATIONS` is an environment variable which, when set, skips all of the migration stuff and just starts PostgreSQL on whatever is currently in `/var/lib/postgresql/`. ZFS backups is just one of the ways how it gets used. Please leave support for that environment variable if you can.

> ^pgmigrate has a --dryrun option, which rolls back rather than committing, but it doesn't seem to actually log the SQL anywhere when that flag is set, so I'm not sure of the best way to output the pending SQL code without running it. I think, though, that all it's going to do each time is run the files in /migrations sequentially?

That's unfortunate. Let's skip it for now then.

> I'm considering the best way to run pg_dump to get the reference schema file. Based on the permissions errors I've been getting when attempting to run as root and postgres, as well as a read of the Postgres docs, it seems like I need to execute the command as the mediacloud superuser. When I've tried that (specifically, executing pg_dump --dbname=mediacloud --username=mediacloud in initialize_db.sh), I get the error: pg_dump: [archiver (db)] connection to database "mediacloud" failed: FATAL: Peer authentication failed for user "mediacloud". Any thoughts on how to solve this one?

PostgreSQL auth is a confusing mix between UNIX accounts and database's own ones so don't feel bad. Currently implemented `postgresql-server` is configured to authenticate as `mediacloud` via [environment variables](https://github.com/mediacloud/backend/blob/88618c545efbb9b9744ea89d18fe413d7a3a5bb4/apps/postgresql-server/Dockerfile#L53-L57) so you should be able to do something like:

```bash
docker run -it dockermediacloud/postgresql-server    # defaults to ":latest" tag
```

and then

```bash
docker exec -it <container_id> psql
```

So dunno, maybe run `pg_dump` at some point when building and then add a shell script in `./dev/` that would `cat` the schema from `postgresql-server:latest`?

.gitignore
```diff
@@ -25,6 +25,9 @@ nytprof*
 # Devel::Cover output
 cover_db/
 
+# Schema file dumped for reference in development
+**/mediawords.sql
```
jtotoole, line 29: I figured it could create confusion if this is being committed by different people during overlapping development, and the file isn't used anywhere at this point

jtotoole:
not sure how to go about testing this in production and extremely paranoid about doing so, but i think the PR is in decent shape now, and behaving as intended locally

pypt:
Thanks, will have a look soon.

.gitignore
pypt: Good idea, thanks!

apps/postgresql-server/Dockerfile
```diff
@@ -10,24 +10,44 @@ RUN \
...
+    apt-get -y --no-install-recommends install build-essential python3-dev libffi-dev libpq-dev && \
+    #
+    # Install package to manage schema migrations
+    pip3 install yandex-pgmigrate==1.0.6 && \
+    #
+    # Remove temporary dependencies
+    apt-get -y remove build-essential python3-dev libffi-dev libpq-dev && \
+    #
+    # Cleanup
+    apt-get -y autoremove && \
```
pypt, line 29: Add an extra:

```shell
apt-get -y clean
```

here too if you can. It [removes downloaded APT packages](https://www.networkworld.com/article/3453032/cleaning-up-with-apt-get.html).

apps/postgresql-server/Dockerfile
```diff
@@ -10,24 +10,44 @@ RUN \
...
 
-# Initialize data volume, create users, a database, and initialize it with
-# schema
+# Copy helper scripts, schema, migrations, pgmigrate callbacks/config
+RUN mkdir -p /opt/mediacloud/
+COPY bin/ /opt/mediacloud/bin/
+COPY pgmigrate-callbacks/ /opt/mediacloud/pgmigrate-callbacks/
+COPY migrations/ /opt/mediacloud/migrations/
+COPY migrations.yml /opt/mediacloud/migrations.yml
+RUN cd /opt/mediacloud
```
pypt, line 44: I don't think you need to `cd` anywhere here at all, but in any case it [doesn't do what it you intended for it to do](https://stackoverflow.com/a/58848691) - you'd need `WORKDIR` for that.

apps/postgresql-server/bin/apply_migrations.sh
```diff
@@ -55,67 +29,8 @@ echo "New schema version: ${NEW_SCHEMA_VERSION}"
...
-
-    # Commit transaction
-    echo "COMMIT;" >> "${CONCAT_MIGRATION}"
-
-    # Apply migration
-    psql -v ON_ERROR_STOP=1 -p "${TEMP_PORT}" -f "${CONCAT_MIGRATION}"
-
-    rm "${CONCAT_MIGRATION}"
-fi
+# apply migrations
```
pypt, line 32: It doesn't seem to work on second run of the container when the migrations *do* have to be applied with the server running on a temporary port before restarting it on the public (5432) port:

```bash
➜ docker run -it gcr.io/mcback/postgresql-server:jot-pgmigrate bash
postgres@4250ea85468a:/$ /opt/postgresql-server/bin/postgresql.sh
Skipping schema migrations on first run...
2021-06-08 08:15:58 EDT [9-1] LOG:  starting PostgreSQL 13.3 (Ubuntu 13.3-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit
2021-06-08 08:15:58 EDT [9-2] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2021-06-08 08:15:58 EDT [9-3] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
2021-06-08 08:15:58 EDT [27-1] LOG:  database system was shut down at 2021-06-08 08:15:31 EDT
2021-06-08 08:15:58 EDT [9-4] LOG:  database system is ready to accept connections
^C2021-06-08 08:16:02 EDT [9-5] LOG:  received fast shutdown request
2021-06-08 08:16:02 EDT [9-6] LOG:  aborting any active transactions
2021-06-08 08:16:02 EDT [9-7] LOG:  background worker "logical replication launcher" (PID 33) exited with exit code 1
2021-06-08 08:16:02 EDT [28-1] LOG:  shutting down
2021-06-08 08:16:03 EDT [9-8] LOG:  database system is shut down

postgres@4250ea85468a:/$ /opt/postgresql-server/bin/postgresql.sh
Applying schema migrations...
waiting for server to start....2021-06-08 08:16:04 EDT [47-1] LOG:  starting PostgreSQL 13.3 (Ubuntu 13.3-1.pgdg20.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0, 64-bit
2021-06-08 08:16:04 EDT [47-2] LOG:  listening on IPv4 address "0.0.0.0", port 12345
2021-06-08 08:16:04 EDT [47-3] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.12345"
2021-06-08 08:16:04 EDT [48-1] LOG:  database system was shut down at 2021-06-08 08:16:02 EDT
2021-06-08 08:16:04 EDT [47-4] LOG:  database system is ready to accept connections
 done
server started
Traceback (most recent call last):
  File "/usr/local/bin/pgmigrate", line 8, in <module>
    sys.exit(_main())
  File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 889, in _main
    config = get_config(args.base_dir, args)
  File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 816, in get_config
    conf = conf._replace(conn_instance=_create_connection(conf))
  File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 171, in _create_connection
    conn = _create_raw_connection(
  File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 163, in _create_raw_connection
    conn = psycopg2.connect(conn_string, connection_factory=LoggingConnection)
  File "/usr/local/lib/python3.8/dist-packages/psycopg2/__init__.py", line 127, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Cannot assign requested address
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?

postgres@4250ea85468a:/$ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
postgres       1  0.0  0.0   7420  3776 pts/0    Ss   08:15   0:00 bash
postgres      47  0.0  1.7 5836140 277276 ?      Ss   08:16   0:00 /usr/lib/postgresql/13/bin/postgres -D /var/lib/postgresql/13/main -c config_file=/etc/po
postgres      49  0.0  0.0 5836140 4196 ?        Ss   08:16   0:00 postgres: 13/main: checkpointer 
postgres      50  0.0  0.3 5836304 55752 ?       Ss   08:16   0:00 postgres: 13/main: background writer 
postgres      51  0.0  0.1 5836140 22112 ?       Ss   08:16   0:00 postgres: 13/main: walwriter 
postgres      52  0.0  0.0 5837116 8616 ?        Ss   08:16   0:00 postgres: 13/main: autovacuum launcher 
postgres      53  0.0  0.0  70584  5516 ?        Ss   08:16   0:00 postgres: 13/main: stats collector 
postgres      54  0.0  0.0 5836952 6496 ?        Ss   08:16   0:00 postgres: 13/main: logical replication launcher 
postgres      62  0.0  0.0   9208  3228 pts/0    R+   08:18   0:00 ps aux
```

I think the port is misconfigured here (maybe).

apps/postgresql-server/bin/initialize_db.sh
```diff
@@ -35,8 +35,11 @@ CREATE DATABASE mediacloud WITH
 EOF
 psql -v ON_ERROR_STOP=1 -c "${CREATE_DB_SQL}"
 
-# Initialize with schema
-psql -v ON_ERROR_STOP=1 -d mediacloud -f /opt/postgresql-server/schema/mediawords.sql
+# Run migrations with pgmigrate package
+cd /opt/mediacloud && pgmigrate -t latest migrate
+
+# Dump schema file for reference in development
+psql -v ON_ERROR_STOP=1 mediacloud -c '\! pg_dump mediacloud > /tmp/mediawords.sql'
```
pypt, line 42: 1. `/tmp` and `/var/tmp` could be tmpfs filesystems mounted by Docker to the container; I think it's better to store the generated schema somewhere less temporary, e.g. `/`;
2. Instead of running `pg_dump` from within `psql` (`psql`'s `\!` command just starts a shell), you can just run `pg_dump` directly;
3. By the way, by default `psql` just ignores errors that it encounters in the input SQL. For example, if you had the following SQL file:

    ```sql
    CREATE TABLE foo (name TEXT);
    
    blergh;
    
    CREATE TABLE bar (name TEXT);
    ```

    and were to run `psql -d database_name -f that_file_with_a_typo_in_the_middle.sql`, it would `CREATE TABLE foo`, complain about the `blergh` statement and then happily `CREATE TABLE bar`. This is something to be wary of when, for example, importing large dumps because one might end up with an incomplete imported dump.

dev/get_schema.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+#
+# Generate latest schema file for reference in development 
+# To be run from project root (so, './dev/get_schema.sh')
```
pypt, line 4: Can you make it work from any directory? You can find out where the script is located and then point the schema to `$PWD/../apps/`, similarly to how it's done here:

https://github.com/mediacloud/backend/blob/d1da1bebba9430693ee9a32d0daa3f717b53dbb1/apps/nytlabels-annotator/src/crappy-predict-news-labels/nytlabels.sh#L6-L8

dev/get_schema.sh
```diff
@@ -0,0 +1,11 @@
...
+#
+
+set -u
+set -e
+
+docker pull dockermediacloud/postgresql-server:master
+docker run dockermediacloud/postgresql-server:master cat /tmp/mediawords.sql > ./apps/postgresql-server/schema/mediawords.sql
```
pypt, line 11: This is now `gcr.io/mcback`, not `dockermediacloud` anymore. Forgot to change it on my merge, sorry!

Also, I wouldn't pull anything here and would use the `latest` Docker tag instead of `master`. `latest` always [points (should point) to the last built / pulled image](https://github.com/mediacloud/backend/blob/master/doc/creating_apps.markdown#tag-versions), so if one is running this dev script, they're expecting to get the schema from the latest built / pulled image (be it `master`, `release` or some sort of a feature branch).

pypt (changes requested):
Thank you for all of your work on this, and sorry for the 428342837th time for the delay.

Just some minor changes here and there, plus a single bug (migrations don't seem to work on second run of the database service in the container).

dev/get_schema.sh
jtotoole: Running into a funny problem when using `latest` tag rather than `master`:

```
jamesotoole@Jamess-MacBook-Pro-2 dev (jot-pgmigrate) $ ./get_schema.sh
Using default tag: latest
Error response from daemon: manifest for gcr.io/mcback/postgresql-server:latest not found: manifest unknown: Failed to fetch "latest" from request "/v2/mcback/postgresql-server/manifests/latest".
```
Confusing because I thought Docker automatically appends the latest tag to the most recently built image

apps/postgresql-server/bin/apply_migrations.sh
jtotoole: Ah right you are, I had to pass a connection string argument with `port=12345` to override the default `5432`.

jtotoole:
Alrighty, I think this is working correctly now save for my confusion re: the `latest` tag in `get_schema.sh`.

dev/get_schema.sh
pypt: > Confusing because I thought Docker automatically appends the latest tag to the most recently built image

It does, but it doesn't (and shouldn't) push it to the Docker repository - `latest` is just, uh, a marker on your own machine pointing to the latest image that you yourself have pulled or built. Given that you don't (and shouldn't) `docker push ...:latest` to the repository, the `:latest` can't be pulled from it either.

Here's a bit more on that:

https://medium.com/@mccode/the-misunderstood-docker-tag-latest-af3babfd6375

And again, our own doc:

https://github.com/mediacloud/backend/blob/master/doc/creating_apps.markdown#tag-versions

So, simply don't `docker pull` anything here (as you want to get the schema of whatever was last pulled / built locally), and use the `latest` tag instead of `master`, `release` or anything else.

pypt:
Thanks James, I'll have a look soon!

Why are all the tests failing though?

apps/postgresql-server/bin/postgresql.sh
```diff
@@ -19,5 +19,8 @@ else
     echo "Done applying schema migrations."
 fi
 
+# dump schema file for reference in development (run ./dev/get_schema.sh to get local copy)
+pg_dump mediacloud > /opt/postgresql-server/schema/mediawords.sql
+
```
pypt, line 24: This here has to be done once, at Docker build time, not in the PostgreSQL service startup wrapper, otherwise it will attempt to build the whole 16 TB worth of data on every PostgreSQL start :)

jtotoole: Back to `initialize_db.sh` it goes 🙂 

This is the result of me throwing the kitchen sink at this PR because of a funny problem wherein the GitHub Actions machine for some reason fails to create the `postgresql-server/schema` directory, despite successfully creating the other subdirectories specified in [the same `Dockerfile` command](https://github.com/mediacloud/backend/pull/759/files#diff-97559d4961873beb9ff7b69e99dbaff753262a2a30deb9790dbe8344689000c7R33-R37) (`bin`, `migrations`, `pgmigrate-callbacks`). Illustrated here via [tmate debugging](https://github.com/mediacloud/backend/pull/759/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R193-R197) (i.e. `ssh`ing to the box running the GHA build):

```
runner@fv-az118-420:~/work/backend/backend/apps/postgresql-server$ ls
Dockerfile  bin  migrations  migrations.yml  pgmigrate-callbacks
```

Consequently, the GHA builds are [failing](https://github.com/mediacloud/backend/pull/759/checks?check_run_id=2880302777) with the error `Cannot create container for service postgresql-server: invalid mount config for type "bind": bind source path does not exist: /home/runner/work/backend/backend/apps/postgresql-server/schema`.

When I run the build and create the container locally, I face no such problem and the tests pass:

```
postgres@b2be31a6edca:/opt$ cd postgresql-server
postgres@b2be31a6edca:/opt/postgresql-server$ ls
bin  migrations  migrations.yml  pgmigrate-callbacks  schema
postgres@b2be31a6edca:/opt/postgresql-server$ cd schema
postgres@b2be31a6edca:/opt/postgresql-server/schema$ ls
mediawords.sql
```
Do you have any hypotheses about the cause of this disconnect between build contexts?

pypt: In `docker-compose.tests.yml` we try to mount as source code from the host to the container, on top of what got baked in via `Dockerfile`; the idea is that if the source code / configuration / whatever gets changed on the host (e.g. when you're actively developing things and trying to make something work), you wouldn't have to rebuild images of affected apps after each and every change.

We do this not only for the "app" at hand but for its dependencies too, e.g. one can write a migration and run a test straight away without having to rebuild `postgresql-server` (because the migration would get applied on the container start).

Therefore, in `docker-compose.tests.yml` we mount the `apps/postgresql-server/bin/`, `apps/postgresql-server/schema/` (which used to include `migrations/`), and `apps/postgresql-base/conf/`:

https://github.com/mediacloud/backend/blob/133ae634d7809c8e6fdbcb9931d335d2b4420701/apps/common/docker-compose.tests.yml#L86-L101

I think the problem that you're encountering here is that the `schema/` directory (rightfully) gets removed on this branch, and Git doesn't store information about directories (only about files):

```bash
$ git status
On branch jot-pgmigrate
Your branch is up to date with 'origin/jot-pgmigrate'.
nothing to commit, working tree clean

$ mkdir git_doesnt_care_about_directories

$ git status                             
On branch jot-pgmigrate
Your branch is up to date with 'origin/jot-pgmigrate'.
nothing to commit, working tree clean
```

so while an empty `schema/` directory exists on your laptop, `schema/` is nowhere to be found right after checking out the repository on the GitHub actions worker. Then `docker-compose.tests.yml` of various apps try to mount that `schema/` directory but it doesn't exist (and shouldn't), and so all of the tests fail. When you resurrected `apps/postgresql-server/schema/migrations.sql`, `apps/postgresql-server/schema/` started existing again, and tests started passing again :)

Given that `schema/` is now gone, you should update (search-and-replace basically) instances of `postgresql-server` in all of the `docker-compose.tests.yml`. I think that for that service it would be sensible to mount the following things from the host:

* `postgresql-server/bin/` - PostgreSQL server startup wrapper script
* `postgresql-server/migrations/` - new migrations location
* `postgresql-base/conf/` - PostgreSQL configuration

Also get rid of `postgresql-server/schema/mediawords.sql` from the source tree, it's better off in root (`/`).

It's not applicable here, but in the future if you'd like to create an empty ignored directory in a Git repository, one hack to do it is to add a `.gitignore` like this one:

https://github.com/mediacloud/backend/blob/133ae634d7809c8e6fdbcb9931d335d2b4420701/apps/solr-base/src/solr/collections/mediacloud/data/.gitignore

.github/workflows/build.yml
```diff
@@ -189,3 +189,16 @@ jobs:
...
+        if: always()
+      
+          # - name: Setup tmate session
+      #   uses: mxschmitt/action-tmate@v3
+      #   if: always()
+      #   with:
+      #     limit-access-to-actor: true
+
+      # ^uncomment this if you need to debug the GitHub actions build
+      # TODO: link to new README section once the related branch is merged
```
pypt, line 204: Can you revert this before merging? This is a part of https://github.com/mediacloud/backend/pull/765 I think.

pypt (changes requested):
A single tiny revert please, plus update a bunch of `docker-compose.tests.yml`.

.github/workflows/build.yml
jtotoole: for sure, i didn't intend to leave it in for the final PR—still trying to work out the issue of the GHA build not working

jtotoole:
Looks like the latest failures are all crimson hexagon related (so no more volume problems I don't think)—I can confirm once https://github.com/mediacloud/backend/pull/793 is merged

jtotoole:
@pypt alrighty, only test failures are crimson hexagon-related, so i think this is good to go assuming it looks okay to you

apps/common/docker-compose.tests.yml
```diff
@@ -94,8 +94,8 @@ services:
               source: ./../postgresql-server/bin/
               target: /opt/mediacloud/bin/
             - type: bind
-              source: ./../postgresql-server/schema/
-              target: /opt/mediacloud/schema/
+              source: ./../postgresql-server/pgmigrate/
+              target: /opt/mediacloud/pgmigrate/
```
pypt, line 98: Here and elsewhere, the mount point is wrong - given that you bake in `pgmigrate` directory to `/opt/postgresql-server/`:

https://github.com/mediacloud/backend/blob/a69189c18e723ff5b1c3c43a1bac42fd10bd4900/apps/postgresql-server/Dockerfile#L41

you should be mounting `postgresql-server/pgmigrate/` from the host to `/opt/postgresql-server/` (not `/opt/mediacloud/`) on the guest too, so that whenever `postgresql-server` service gets started from any of the `docker-compose.tests.yml`, it "sees" the latest `pgmigrate` directory from the host and not whatever is baked into the image.

pypt (changes requested):
One more quick find-and-replace, and we're good to go!

pypt:
Amazinglymazing, thank you so much!

pypt:
Fixes #754.

esirK:
Hey @jtotoole  @pypt 
I restarted our Swarm and since it's using `gcr.io/mcback/postgresql-server:release` for the PostgreSQL image, it picked up all the latest changes which I think might be breaking the system. I'm currently getting the following error
```
psycopg2.errors.DuplicateTable: relation "database_variables" already exists

 File "/usr/local/bin/pgmigrate", line 8, in <module>
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     sys.exit(_main())
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |   File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 891, in _main
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     COMMANDS[args.cmd](config)
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |   File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 753, in migrate
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     _migrate_step(state, config.callbacks, config.base_dir, config.user,
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |   File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 555, in _migrate_step
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     _apply_version(version, base_dir, user, schema, cursor)
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |   File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 465, in _apply_version
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     _apply_file(version_info.file_path, cursor)
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |   File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 451, in _apply_file
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     _apply_statement(statement, cursor)
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |   File "/usr/local/lib/python3.8/dist-packages/pgmigrate.py", line 442, in _apply_statement
cfamediacloud_postgresql-server.1.nmvjhex3xf5b@ip-172-31-10-89    |
cfamediacloud_postgresql-server.1.nmvjhex3xf5b@ip-172-31-10-89    | During handling of the above exception, another exception occurred:
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    |     raise MigrateError('Unable to apply statement')
cfamediacloud_postgresql-server.1.pq7qkzbi9nxr@ip-172-31-10-89    | pgmigrate.MigrateError: Unable to apply statement
cfamediacloud_postgresql-server.1.nmvjhex3xf5b@ip-172-31-10-89    |
```

any suggestions on how I can resolve this?

esirK:
>MC_POSTGRESQL_SKIP_MIGRATIONS is an environment variable which, when set, skips all of the migration stuff and just starts PostgreSQL on whatever is currently in /var/lib/postgresql/. ZFS backups is just one of the ways how it gets used. Please leave support for that environment variable if you can.

From the comments, I see I can set `MC_POSTGRESQL_SKIP_MIGRATIONS` as an environment variable. Where should I set this from?

PR #762, opened 2021-01-27
fix regex for google podcast test

jtotoole:
Another bite at the apple: Regex wasn't working quite as intended in [the previous PR](https://github.com/mediacloud/backend/pull/752) because I was [using a wildcard for the integer after the podcast feed url](https://github.com/mediacloud/backend/pull/752/files#diff-23e631c3d080992c4b1e9a326393af0edd28348de776468bcf2dc12e511d7b61R127) along with `.*` for a wildcard string of unknown length. Turns out there's an alternate feed link a bit further down that's appended with `;3`, and so [`match.group(1)`](https://github.com/mediacloud/backend/blob/ae79a48362b3ef33a31c89e60258685097e145bc/apps/crawler-fetcher/src/python/crawler_fetcher/handlers/feed_podcast.py#L132) was grabbing that instead.

When we instead [hard-code a `2`](https://github.com/mediacloud/backend/pull/762/files#diff-23e631c3d080992c4b1e9a326393af0edd28348de776468bcf2dc12e511d7b61R127) in that regex, it works correctly for the main feed on the NPR podcast page as well as 5 other show pages I looked at.

### DIFF
--- a/apps/crawler-fetcher/src/python/crawler_fetcher/handlers/feed_podcast.py
+++ b/apps/crawler-fetcher/src/python/crawler_fetcher/handlers/feed_podcast.py
@@ -124,7 +124,7 @@ def _get_feed_url_from_google_podcasts_url(url: str) -> str:
             html = res.decoded_content()
 
     # get show's feed URL from its Google Podcasts homepage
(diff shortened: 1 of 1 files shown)


pypt:
Thanks, seems to work now!

PR #763, opened 2021-02-01
Host Docker images on Google Container Registry instead of Docker Hub

pypt:
We've been hosting our pre-build Docker images in [Docker Hub, under `dockermediacloud` username](https://hub.docker.com/u/dockermediacloud).

Recently Docker Hub introduced [rate limiting for public images](https://www.docker.com/increase-rate-limits) (such as ours) without much of a heads-up, and I messed up a couple of deployments because Docker utility didn't let me pull the images that make up Media Cloud backend.

Paying $35 for them to lift the pull cap would be fine, but a slightly bigger inconvenience is that a subscription to their service still prevents images from being pulled when the client that's pulling said images is not logged in via `docker login`. Maintaining login credentials on all servers is cumbersome, so I figured I'll just move everything to [Google Cloud's Container Registry](https://cloud.google.com/container-registry) instead. Not only it does not require login, but it should be cheaper too (I estimate that we'll be spending up to $10/mo for our container needs).

Thus, image naming changes from:

`dockermediacloud/<app_name>:tag`

to:

`gcr.io/mcback/<app_name>:tag`

I've added a little doc on how to create service accounts for CI builds to push Docker images to GCR.

In addition to reviewing the PR, could you also verify that you yourself can push something to GCR using your `@mediacloud.org` account? To do that, you'll need to [install Google Cloud SDK](https://cloud.google.com/sdk/docs/install), run `gcloud auth login` (make sure to log in with the right Google account), run `gcloud config set project mcback`, pull some random image from docker and `docker tag` it to something like `gcr.io/mcback/test:test`, and then `docker push` the image to GCR.


jtotoole:
Nicely done!

Pulling works, but I can't push due to a permissions error (`unauthorized: You don't have the needed permissions to perform this operation, and you may have invalid credentials`). 

I'm already logged in with my mediacloud account, so I'm guessing I need to be adding as an editor on the `mcback` project?

pypt:
Oh, and pushes to GCR are faster too (21:45 vs 11:31).

pypt:
Merging as I'll want to deploy some stuff soon.

@jtotoole, could you try pushing again?

jtotoole:
success https://console.cloud.google.com/gcr/images/mcback/GLOBAL/jot-test?project=mcback&authuser=1&gcrImageListsize=30

pypt:
Great, thanks!

PR #765, opened 2021-02-04
[WIP] Print test failures in GH Actions

jtotoole:


pypt:
Is this PR (or the other one with a similar name) something to look at already, or is it still work in progress?

jtotoole:
Not ready for prime time yet; I'm facing a stumbling block that I figured we could discuss in more detail tomorrow. Specifically, the `awk` command [I added](https://github.com/mediacloud/backend/pull/765/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R193) as a new build step to print just the test failures seems [unable to interact properly](https://github.com/mediacloud/backend/pull/765/checks?check_run_id=1834488888) with `joblog.txt.`:

Screen Shot 2021-02-08 at 10 58 23 AM

The [GH Actions logs](https://pipelines.actions.githubusercontent.com/Maij8RrtXQn1PDAb2MdBdli6vOOlymziYGbuxBLl2OAfgGzL0o/_apis/pipelines/1/runs/597/signedlogcontent/22?urlExpires=2021-02-08T14%3A32%3A45.6463996Z&urlSigningMethod=HMACV1&urlSignature=nyMs8KlLl8Ag04RApOpMGOvY1rybDiiVww%2Bbx6qXqiU%3D) have nothing to say on the subject, so I'm at a loss.

Closed [the other PR](https://github.com/mediacloud/backend/pull/764) related to this as I think the approach here is more likely to be useful.

pypt:
I think GHA doesn't run step *n+1* if step *n* has failed, so you might want to either make the "Run tests" step exit with a zero error code, merge the "Run tests" and "Print test failures" into a single step, or (perhaps better approach) [add some conditionals](https://github.community/t/continue-on-error-allow-failure-ui-indication/16773).

Also, running the full test suite takes a whole lot of time so you might consider temporarily removing a bunch of tests from the codebase just to make the GHA run finish faster. Later you can just `git revert` the commit.

jtotoole:
Finally got this working as intended—see the "List errors and test failures" step in [this build](https://github.com/mediacloud/backend/runs/2053564412)

jtotoole:
ready for review!

pypt:
Merged in `master` to trigger a rebuild (the old one has expired).

pypt:
So printing failures as GitHub Actions annotations didn't work?

jtotoole:
I initially abandoned that approach because the GitHub Checks API will only publish annotations on test files that have been changed in the PR. There's [a marketplace action](https://github.com/EnricoMi/publish-unit-test-result-action) that will annotate everywhere if we first create a test summary file in JUnit XML format, but that seemed like rather high-hanging fruit compared to this approach. That said, I'm happy to give the JUnit thing a shot if you think it'd be better and not wildly complicated.

.github/workflows/build.yml
```diff
@@ -189,3 +189,7 @@ jobs:
           MC_TWITTER_ACCESS_TOKEN_SECRET: ${{ secrets.MC_TWITTER_ACCESS_TOKEN_SECRET }}
           MC_TWITTER_CONSUMER_KEY: ${{ secrets.MC_TWITTER_CONSUMER_KEY }}
           MC_TWITTER_CONSUMER_SECRET: ${{ secrets.MC_TWITTER_CONSUMER_SECRET }}
+
+      - name: List test failures and errors
+        run: echo -e 'Test failures:\n' && awk '/FAILURES/,/short/' /home/runner/runners/*/_diag/pages/*.log && echo -e '\n\nErrors:\n' && awk '/non-zero/,/status 1/' /home/runner/runners/*/_diag/pages/*.log
```
pypt, line 194: What's in `/home/runner/runners/*/_diag/pages/*.log`, i.e. how are you finding those errors? Do you perhaps have a sample log file from that directory that we could look at (without various private credentials that might be in this log)?

Grepping (awking?) out failures from a log file is a bit wonky because any test might print a string `FAILURES` at any point for a legitimate reason and still succeed; plus, tests don't have to fail with `status 1` specifically as they can have any non-zero exit status.

pypt (changes requested):
A single question, plus I don't think the build works anymore.

Issue #743, opened 2020-11-25
Merge "cliff-fetch-annotation" and "cliff-update-story-tags" into a single app


pypt:
Hey James!

Once you're done with the CLIFF build fixing and had enough of the turkey, could you **merge `cliff-fetch-annotation` and `cliff-update-story-tags` into a single app**?

We have those two apps, `cliff-fetch-annotation` and `cliff-update-story-tags` that use `cliff-annotator` to detect entities in input text (coming from stories):

1. `cliff-fetch-annotation`:
    1. "listens" for jobs on the `MediaWords::Job::CLIFF::FetchAnnotation` RabbitMQ queue,
    2. takes `stories_id` (ID of a row in `stories` PostgreSQL table) as a parameter,
    3. fetches its full text content from `download_texts` (IIRC),
    4. POSTs it to `cliff-annotator`,
    5. gets back a JSON response,
    6. stores the JSON response in `cliff_annotations` table, and
    7. adds a new job to `MediaWords::Job::CLIFF::UpdateStoryTags` queue for the `cliff-update-story-tags` app to do;
2. `cliff-update-story-tags`:
    1. listens for jobs on the `MediaWords::Job::CLIFF::UpdateStoryTags` RabbitMQ queue,
    2. takes `stories_id` as a parameter as well,
    3. fetches the JSON response with detected entities from `cliff_annotations` table,
    4. parses it,
    5. stores the detected entities as story's tags (in the `tags` table), and
    6. adds a new job to `MediaWords::Job::NYTLabels::FetchAnnotation` job for the similar processing to happen with `nytlabels-*` apps.

This way we end up knowing what entities (persons, geographical locations, etc.) does the story mention and we can query for them using the `tags` table.

Why is it a two-step process, i.e. why does the raw JSON response have to be stored and then fetched again later, you might ask? I don't know. When I was implementing it, I thought I had a good reason to make it that way, I think my reasoning was along the lines of "oh, we might want to store more of the stuff from the raw JSON responses in the database at a later date so we'll want to reprocess those raw responses", but we never needed to do that, and the table that stores those raw responses ended up being rather chonky and takes up precious and expensive SSD disk space:

```
mediacloud=# select pg_size_pretty(pg_relation_size('cliff_annotations'));
 pg_size_pretty 
----------------
 281 GB
(1 row)
```

(`nytlabels-*` does similar stuff, and the table is even bigger - 816 GB - but that's going to be another task to get rid of it too.)

[...]

PR #766, opened 2021-02-08
In /api/v2/stories/cliff, /api/v2/stories/nytlabels, call underlying services directly instead of reading annotations from disk

pypt:
References #743.

Unblocks https://github.com/mediacloud/backend/pull/745, https://github.com/mediacloud/backend/pull/751.

Some other minor changes:

* When both API calls can't / refuse to annotate story's text, returned error messages are a little bit more specific.
* Now that `nytlabels-annotator` is rather "right" (takes up only 500 or so MB of RAM to run), I've added it to `webapp-api`'s `docker-compose.tests.yml` and made the tests actually use the underlying annotator.

The not-so-nice part is that `_annotate_story_ids()` below is a Perl implementation of James's Python subroutine that does the same stuff, but to get rid of it we'll have to rewrite the webapp to Python first.


pypt:
Seems to work, even with GHA using real `cliff-annotator` and `nytlabels-annotator` instead of lightweight mocks.

PR #767, opened 2021-02-08
Store container / host system logs in ELK stack

pypt:
This PR adds the so-called ELK stack (Elasticsearch, Logstash, Kibana) to our app world, except that there's no L in this ELK as Logstash has been replaced by Filebeat, Journalbeat and Auditbeat, but I didn't come up with a better abbreviation.

This is an attempt at solving at least two things:

* Some months ago we had a bit of a hiccup with our security (thanks to my ingenuity!), and when trying to identify who might have accessed our systems during the lifetime of the security hiccup, we found out that it's rather hard to do that as the logs which were supposed to log that stuff were long purged (due to disk space constraints in root partitions on all servers), and the ones that we did have were hard to grep due to their size. So, we've decided that we need to aggregate all the logs from all the (backend) systems that we run in one place and make them searchable.
* Some containers (mostly topic-related ones AFAIK) crash here and there, then the Docker Swarm identifies that the container has stopped, starts a new one on the same / different server to keep the replica count right, and lastly cleans up the crashed container by removing it. This makes it rather hard to find out what exactly happened to the crashed container post mortem. Containers currently get logged to host system's journald (weird reinvention of `/var/log/`) but that's not always helpful too as those logs get trimmed as well to preserve space.

Enter ELK. The thing consists of the following:

* Filebeat - tracks `/var/log/` and `/var/lib/docker/containers/*/*.log` (Docker container logs), posts its findings to Elasticsearch
* Auditbeat - tracks auditd's complaining, uploads it to Elasticsearch
* Journalbeat - tracks journald logs, puts everything to Elasticsearch as well
* Elasticsearch - indexes logs submitted by *beats, makes them searchable
* Kibana - interface to Elasticsearch, lets you click on things, filter for stuff and so on

Generally the questions that this setup should be able to provide are something like these:

* What happened between `date1` and `date2` everywhere?
* What happened between `date1` and `date2` on `serverA`?
* What happened between `date1` and `date2` on `serviceX` (running on multiple servers)?
* What happened to `serviceX` on a given date range?
* What happened to `serviceX` when it was processing inputs with `stories_id = 12345`?
* ...

One more extra feature of the setup: it's pre-configured to snapshot indexes daily and upload them to S3, and also I've made it purge indexes that are older than 30 days, so that way we should be able to get an effectively infinite archive of our logs on S3.

Some minor changes:

* OpenJDK updated from version 12 to 15;
* Added `doc/elk_logging.markdown` to describe initial setup that I didn't manage to automate; this doc could serve as a reference point for how to do various things with ELK;
* Added `/container_cpu_limit.sh` helper script in `base` container which outputs the number of CPU cores that the container is limited to.

@jtotoole, would you be able to try running the thing yourself, click around for a bit and see if you find it figureout-able? After pulling the images, something like this should work:

```bash
# "-m" flag maps exposed ports from the main service (elk-kibana in this case) to your host system
$ ./dev/run.py -m elk-kibana bash
container$ /opt/kibana/bin/kibana.sh
```

You'll need S3 credentials (in the form of environment variables), try our local secure storage (don't wanna use the brandname publicly) for that, look up for "elk". Make sure to use the testing prefix!


jtotoole:
Got the container running with the requisite environment variables and was able to load the Kibana dashboard. The dashboard, however, appears unaware of those `snapshots-test` logs:

Screen Shot 2021-03-09 at 2 43 45 PM

Nothing in [the logging](https://gist.github.com/jtotoole/767764e9ee9a2a90623d3f0d7cf53d88) upon server startup looks amiss, and the environment variables appear as expected when I `echo` them in the container. Any idea where I might be going wrong?

pypt:
Try this doc:

https://github.com/mediacloud/backend/blob/master/doc/elk_logging.markdown

There's nothing in the Dashboard, I think it's the Discover tab that's the most useful one as you can query for arbitrary stuff there.

jtotoole:
^yeah, screenshot above is what i get from going to the link (http://localhost:5601/app/management/kibana/indexPatterns) in that doc. i'm assuming i've got some aspect of the configuration wrong, we can talk about it in our meeting tomorrow

pypt:
Sure!

PR #771, opened 2021-02-18
Update postfix headers

jtotoole:
`Message-Id` + `Date` are now working in my testing. I can't get `List-Unsubscribe` to show up despite having followed the guidance at these links:

https://blog.mailtrap.io/list-unsubscribe-header/
https://www.reddit.com/r/postfix/comments/a8p1m7/custom_header/eccyorb?utm_source=share&utm_medium=web2x&context=3

I've been banging my head against this for a while, so if you've got any ideas about the `List-Unsubscribe` header thing, that'd be great. I'm also fine with just forgetting about it given that our updated messages seem no longer to be flagged as spam.

### DIFF
--- a/apps/common/docker-compose.tests.yml
+++ b/apps/common/docker-compose.tests.yml
@@ -18,7 +18,7 @@ services:
             # emails are more likely to be marked spam if we don't have such a header, so
             # we make the email subject "Delete account and unsubscribe" in 
             # mediawords/util/config/common.py
-            MC_UNSUBSCRIBE_MAILTO_LINK: "support@example.com"
+            MC_EMAIL_UNSUBSCRIBE: "support@example.com"
         volumes:
             - type: bind
               source: ./src/
--- a/apps/common/docker-compose.tests.yml
+++ b/apps/common/docker-compose.tests.yml
@@ -13,6 +13,12 @@ services:
             MC_DOWNLOADS_AMAZON_S3_DIRECTORY_NAME: "${MC_DOWNLOADS_AMAZON_S3_DIRECTORY_NAME}"
             MC_PUBLIC_STORE_TYPE: "postgresql"
             MC_PUBLIC_STORE_SALT: "foo"
+            # Email address to point to in List-Unsubscribe email header.
+            # Technically we don't have a straightforward "unsubscribe" endpoint, but our 
+            # emails are more likely to be marked spam if we don't have such a header, so
+            # we make the email subject "Delete account and unsubscribe" in 
+            # mediawords/util/config/common.py
+            MC_UNSUBSCRIBE_MAILTO_LINK: "support@example.com"
         volumes:
             - type: bind
               source: ./src/
--- a/apps/common/src/python/mediawords/util/config/common.py
+++ b/apps/common/src/python/mediawords/util/config/common.py
@@ -155,6 +155,14 @@ class SMTPConfig(object):
         """Password."""
         return ''
 
+    @staticmethod
+    def unsubscribe_address() -> str:
+        """Mailto link for email to which unsubscribe/account deletion requests should be sent"""
+        address = env_value('MC_UNSUBSCRIBE_MAILTO_LINK', required=False, allow_empty_string=True)
+        if address is None or '@' not in address:
+            address = 'support@example.com'
+        return f'mailto:{address}?subject=Delete%20account%20and%20unsubscribe'
+
 
 class DownloadStorageConfig(object):
     """Download storage configuration."""
--- a/apps/common/src/python/mediawords/util/config/common.py
+++ b/apps/common/src/python/mediawords/util/config/common.py
@@ -158,10 +158,11 @@ class SMTPConfig(object):
     @staticmethod
     def unsubscribe_address() -> str:
         """Mailto link for email to which unsubscribe/account deletion requests should be sent"""
-        address = env_value('MC_UNSUBSCRIBE_MAILTO_LINK', required=False, allow_empty_string=True)
+        address = env_value('MC_EMAIL_UNSUBSCRIBE', required=False, allow_empty_string=True)
         if address is None or '@' not in address:
             address = 'support@example.com'
-        return f'mailto:{address}?subject=Delete%20account%20and%20unsubscribe'
+        return address
+        # return f'mailto:{address}?subject=Delete%20account%20and%20unsubscribe'
 
 
 class DownloadStorageConfig(object):
--- a/apps/common/src/python/mediawords/util/config/common.py
+++ b/apps/common/src/python/mediawords/util/config/common.py
@@ -157,12 +157,11 @@ class SMTPConfig(object):
 
     @staticmethod
     def unsubscribe_address() -> str:
-        """Mailto link for email to which unsubscribe/account deletion requests should be sent"""
+        """Email to which unsubscribe/account deletion requests should be sent"""
         address = env_value('MC_EMAIL_UNSUBSCRIBE', required=False, allow_empty_string=True)
         if address is None or '@' not in address:
             address = 'support@example.com'
         return address
-        # return f'mailto:{address}?subject=Delete%20account%20and%20unsubscribe'
 
 
 class DownloadStorageConfig(object):
--- a/apps/common/src/python/mediawords/util/mail.py
+++ b/apps/common/src/python/mediawords/util/mail.py
@@ -117,11 +117,14 @@ def send_email(message: Message) -> bool:
             message_part = MIMEText(message.text_body, 'plain', 'utf-8')
             mime_message.attach(message_part)
 
+        mime_message.add_header('List-Unsubscribe',
+                                'mailto:support@mediacloud.org?subject=Delete%20account%20and%20unsubscribe')
+
         # HTML gets attached last, thus making it a preferred part as per RFC
         if message.html_body:
             message_part = MIMEText(message.html_body, 'html', 'utf-8')
             mime_message.attach(message_part)
-
+        print(mime_message)
         if test_mode_is_enabled():
             log.info("Test mode is enabled, not actually sending any email.")
             log.debug("Omitted email:\n\n%s" % mime_message.as_string())
@@ -136,6 +139,7 @@ def send_email(message: Message) -> bool:
 
             # Send message
             refused_recipients = smtp.sendmail(mime_message['From'], mime_message['To'], mime_message.as_string())
+
             if len(refused_recipients):
                 log.warning("Unable to send email to the following recipients: %s" % str(refused_recipients))
 
@@ -160,4 +164,5 @@ def send_text_email(to: str, subject: str, body: str) -> bool:
     body = decode_object_from_bytes_if_needed(body)
 
     message = Message(to=to, subject=subject, text_body=body)
+
     return send_email(message)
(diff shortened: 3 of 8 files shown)


pypt:
Is this ready for review? I see the CI failing for whatever reason.

apps/topics-map/Dockerfile
```diff
@@ -5,7 +5,10 @@
 FROM gcr.io/mcback/common:latest
 
 # Install Java
-RUN apt-get -y --no-install-recommends install openjdk-8-jre-headless
+RUN \ 
+    apt-get update && \
```
jtotoole, line 9: fix for https://github.com/mediacloud/backend/pull/771/checks?check_run_id=1937604900#step:13:2904

jtotoole:
Headers looking good now:

Screen Shot 2021-02-19 at 3 11 32 PM

jtotoole:
The failing test here is related to Brandwatch:

https://github.com/mediacloud/backend/pull/771/checks?check_run_id=1938298020#step:16:2177

I don't think it's related to any of my changes, but I can take a stab at fixing it before merging if you'd like.

apps/common/src/python/mediawords/util/mail.py
```diff
@@ -117,6 +117,9 @@ def send_email(message: Message) -> bool:
             message_part = MIMEText(message.text_body, 'plain', 'utf-8')
             mime_message.attach(message_part)
 
+        mime_message.add_header('List-Unsubscribe',
```
pypt, line 120: Could you add a quick configuration parameter to make the email address in `List-Unsubscribe` configurable? We used to be the sole user of Media Cloud's backend so back in the day we would just hardcode our emails here and there (and they're still hardcoded at some places), but now Code for Africa people are running an instance, and I think there are at least two more backend instances running out there in the wild. So, we don't want to be getting their unsubscribe emails.

We currently store all (I think) of our configuration in environment variables (for better or for worse). We also try our best to keep that fact isolated from the rest of the system, i.e. instead of reading the required configuration parameter directly via, say, `os.environ`, we encapsulate the configuration either in "common" configuration (available in `mediawords.util.config.common`), or in a per-app configuration (here's a [sample Facebook client configuration](https://github.com/mediacloud/backend/blob/master/apps/facebook-fetch-story-stats/src/python/facebook_fetch_story_stats/config.py)), the reasons being:

* When you have a class between the configuration user and the environment variable, you get a chance to process the env. variable a bit, e.g. cast a number to `int`, cast `0` / `1` to `bool`, parse some sort of an URI, do validation, complain when the env. variable is not set, etc.
* Storing configuration (including secrets) in environment variables is not terrible but not great either, so if we ever come up with a better way to do this (Docker secrets? etcd or something?), we won't have to fish out all usages of environment variables in the whole Python codebase.

So that's why you can't just do `os.environ` :)

To add a new "common" (i.e. potentially used by all "apps") configuration parameter, you'll have to:

* Add a new method to [`SMTPConfig`](https://github.com/mediacloud/backend/blob/17bae7d31b294f92710387978e3cb6ed89df54bd/apps/common/src/python/mediawords/util/config/common.py#L126) (or some other class if that makes more sense to you) that would read the environment variable with your "send unsubscribe emails to this address" value
    * Consider making use of [`env_value()`](https://github.com/mediacloud/backend/blob/17bae7d31b294f92710387978e3cb6ed89df54bd/apps/common/src/python/mediawords/util/config/__init__.py#L18) to save up on some boilerplate for testing whether the env. variable exists, etc.
    * Make it an optional configuration parameter (i.e. if it doesn't get set, things don't crash) but make the default email address to a nonexistent email inbox (assuming that our partners will forget to set this property, we don't want their emails defaulting to our `support@`)
    * Optionally check whether what gets set looks like an email address by verifying that it has `@` in it (there are various ways for testing email address validity but there's no point bothering ourselves with that here as this is an internal configuration variable)
* Whenever you want to use it, just import `SMTPConfig` and call the static method that you've just created to get the email address

The email subject can stay non-configurable.

pypt (changes requested):
Thank you James, nitpicked this a little bit!

apps/common/src/python/mediawords/util/mail.py
pypt: Oh yeah, last step: you'll have to add the newly added configuration variable to both `apps/docker-compose.dist.yml` and `apps/common/docker-compose.tests.yml`. The `docker-compose.dist.yml` entry should have a comment for what this env. variable is there for and have a sample value (possibly commented out), and the `docker-compose.tests.yml` might have a similar commented out entry just to let the user (the future us basically) know that such an configuration environment variable exists and that it's supposed to do something.

I might be sounding extra nitpicky, but I think it's a good practice to keep a record of how each application is supposed to be configured.

apps/mail-postfix-server/header_checks
```diff
@@ -1,5 +1,4 @@
 /^Received:.*with ESMTP /         IGNORE
 /^X-Originating-IP:/    IGNORE
 /^X-Mailer:/            IGNORE
-/^Mime-Version:/        IGNORE
```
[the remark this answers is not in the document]
jtotoole: Correct, that was flagged in the spam tester

apps/topics-map/Dockerfile
```diff
@@ -5,7 +5,10 @@
 FROM gcr.io/mcback/common:latest
 
 # Install Java
-RUN apt-get -y --no-install-recommends install openjdk-8-jre-headless
+RUN \ 
+    apt-get update && \ 
+    apt-get -y --no-install-recommends install openjdk-8-jre-headless && \
+    true
```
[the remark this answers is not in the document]
jtotoole: useful background, appreciate the thoughtful write-up!

jtotoole:
alrighty, i think all set (save for adding the `MC_UNSUBSCRIBE_MAILTO_LINK` var in production)

apps/common/docker-compose.tests.yml
```diff
@@ -13,6 +13,12 @@ services:
             MC_DOWNLOADS_AMAZON_S3_DIRECTORY_NAME: "${MC_DOWNLOADS_AMAZON_S3_DIRECTORY_NAME}"
             MC_PUBLIC_STORE_TYPE: "postgresql"
             MC_PUBLIC_STORE_SALT: "foo"
+            # Email address to point to in List-Unsubscribe email header.
+            # Technically we don't have a straightforward "unsubscribe" endpoint, but our 
+            # emails are more likely to be marked spam if we don't have such a header, so
+            # we make the email subject "Delete account and unsubscribe" in 
+            # mediawords/util/config/common.py
+            MC_UNSUBSCRIBE_MAILTO_LINK: "support@example.com"
```
pypt, line 21: Could you rename `MC_UNSUBSCRIBE_MAILTO_LINK` to something like `MC_EMAIL_FROM_ADDRESS` or `MC_MAIL_FROM_ADDRESS`?

* The rest of the email-related settings are prefixed with either `MC_EMAIL_` or `MC_MAIL_`
    * It's not great that we ended up with both `MC_EMAIL_` and `MC_MAIL_` env. var. prefixes, I think it was me who messed it up, but I guess fixing that is out of scope
* The env. variable getter is called `unsubscribe_address()`
* `_MAILTO_LINK` would kind of imply that instead of just the address (`support@example.com`) the variable is supposed to have the whole `mailto:` link in it (`mailto:support@example.com`), and the env. variable getter wouldn't test for that, so one could accidentally end up with a final header resulting into something like `List-Unsubscribe: mailto:mailto:support@example.com` which, depending on how attentive the receiving server is, could lead to emails being rejected due to an invalid header

apps/common/src/python/mediawords/util/config/common.py
```diff
@@ -155,6 +155,14 @@ def password() -> str:
         """Password."""
         return ''
 
+    @staticmethod
+    def unsubscribe_address() -> str:
```
pypt, line 159: Here despite the method being called `unsubscribe_address()` (if I read that I would expect it to return just the `support@example.com` email address) you're returning the full `mailto:` link including the proposed subject line.

I would make the `unsubscribe_address()` method return *just* the address that the unsubscribe emails should be sent to, and then let the caller (`send_email()`) decide where to put this address and what subject does it want to append to it (if any) as this is what a method (or rather the name of it) promises.

apps/mail-postfix-server/header_checks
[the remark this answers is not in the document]
pypt: Okay, thanks.

pypt (changes requested):
Just a quick rename / unify naming if you may.

apps/topics-map/Dockerfile
```diff
@@ -5,7 +5,10 @@
 FROM gcr.io/mcback/common:latest
 
 # Install Java
-RUN apt-get -y --no-install-recommends install openjdk-8-jre-headless
+RUN \ 
+    apt-get -y update && \
```
pypt, line 9: Another note - here's why one sometimes needs to run `apt-get update`:

https://github.com/mediacloud/backend/blob/master/doc/docker_issues.markdown#apt-get-no-longer-works-it-tells-me-to-maybe-run-apt-get-update

It's annoying, so let me know if you think of a good solution for this.

jtotoole: maybe a cron to rebuild `base` once a week or something?

apps/common/src/python/mediawords/util/config/common.py
```diff
@@ -155,6 +155,15 @@ def password() -> str:
...
+        """Mailto link for email to which unsubscribe/account deletion requests should be sent"""
```
pypt, line 160: This docstring is now misleading :)

apps/common/src/python/mediawords/util/config/common.py
```diff
@@ -155,6 +155,15 @@ def password() -> str:
...
+        address = env_value('MC_EMAIL_UNSUBSCRIBE', required=False, allow_empty_string=True)
+        if address is None or '@' not in address:
+            address = 'support@example.com'
+        return address
+        # return f'mailto:{address}?subject=Delete%20account%20and%20unsubscribe'
```
pypt, line 165: Could you remove this commented out line?

pypt (changes requested):

apps/topics-map/Dockerfile
pypt: A daily / weekly cron rebuild would work, the job would have to:

* Skip pulling `base` container image when pulling pre-built images to trigger a full rebuild;
* Doing the rest as usual.

If you're up for it, think it's worth it, and are ready to mess around with GitHub Actions a little bit more, you can do it, but that would have to be a separate issue in any case.

jtotoole: Made an issue [here](https://github.com/mediacloud/backend/issues/775) but seems like something to put on the backburner for now while I work on my other tasks.

apps/common/src/python/mediawords/util/config/common.py
jtotoole: 🙈

apps/topics-map/Dockerfile
pypt: Agree on the backburner note.

pypt (approved):

pypt:
All good now, thank you James!

PR #773, opened 2021-03-01
Speed up NYTLabels annotation by running less models, enabling HTTP/1.1

pypt:
I had a bit of an X-Files and the subsequent TIL moment here so I figured I'll share:

On a Friday about a week ago, one of the indexes on the `tags` table (where, in addition to other things, the labels generated by `cliff-annotator` and `nytlabels-annotator` go to) got messed up and lead to duplicate tags ending up in that table. That took me about a week to fix up (I've fished out the duplicate tags, removed them individually (including the references to them), ran a `CREATE INDEX CONCURRENTLY` to build a duplicate index, and lastly removed the old corrupt one).

While I was busy typing away, a backlog of stories built up - given that the `nytlabels-fetch-annotation` service was shut down, nothing was processing entries in the `MediaWords::Job::CLIFF::FetchAnnotation` queue (that's the queue where stories end up right after extracting them - you can have a look at it if you SSH into one of the servers with SSH tunnels set up and open `http://127.0.0.1:15672/`), and that queue grew to 2m or so stories.

When things were made right again, the backlog of stories made their way through the CLIFF queue but then got stuck in the NYTLabels (`MediaWords::Job::NYTLabels::FetchAnnotation`) queue as `nytlabels-annotator` was processing them only at a rate of 0.6 story per second.

Usual suspects (trying to increase the number of workers, [checking for PostgreSQL locks](https://wiki.postgresql.org/wiki/Lock_Monitoring#.D0.A1ombination_of_blocked_and_blocking_activity)) didn't lead anywhere. Weirdly enough, `nytlabels-annotator`, when called directly, was rather snappy but it worked superslowly in production.

One of the things that did actually improve performance was increasing the number of CPUs allowed for the service (8c50bd98cdf864e907b1cf16cd2f26996e29c7bc). It turns out that [`onnxruntime`](https://github.com/microsoft/onnxruntime) uses multiple threads to run the model, and with its threads running on 8 cores instead of just one things are considerably more snappy.

However, the performance was still disappointing. As you can see in this PR, I've tried various things:

* Installed OpenMP (`libgomp1`) into the `nytlabels-annotator` container image so that `onnxruntime` could use it if it chose to;
* Added an option to run only a few models against the input text instead of all of them; you might recall from the "Merge CLIFF and NYTLabels workers" task that we use only the `descriptors600` key from the NYTLabels result, so there's no point in running the text against the rest of the models;
* Messed with `onnxruntime` internals to shave off a millisecond or two to the runtime.

No matter what I did, each and every request to the `nytlabels-annotator` still took more than one second.

The X-Files and TIL moment happened when, after three hours of debugging, I thought of looking into how exactly requests were being made to the service. `nytlabels-annotator` has this short [`README.md` detailing how to test out the service](https://github.com/mediacloud/backend/blob/7435f93aa20b3ccfc34efef432fd59f44deb927f/apps/nytlabels-annotator/README.md). If you go through it and run `curl` with specific flags that show you not only the headers that get sent back / forth but also the timestamps, it makes it easier to spot the delay:

```
$ echo '{}' | \                                                       
  jq --arg key0 text --arg value0 "$(cat test.txt)" '. | .[$key0]=$value0' | \
  curl --verbose --silent --trace-time --header "Content-Type: application/json" -X POST --data-binary @- http://127.0.0.1:8080/predict.json | \
  jq ".descriptors600"

20:03:24.186338 *   Trying 127.0.0.1:8080...
20:03:24.186380 * TCP_NODELAY set
20:03:24.186506 * Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
20:03:24.186549 > POST /predict.json HTTP/1.1
20:03:24.186549 > Host: 127.0.0.1:8080
20:03:24.186549 > User-Agent: curl/7.68.0
20:03:24.186549 > Accept: */*
20:03:24.186549 > Content-Type: application/json
20:03:24.186549 > Content-Length: 3774
20:03:24.186549 > Expect: 100-continue
20:03:24.186549 > 
20:03:25.188675 * Done waiting for 100-continue
20:03:25.188986 } [3774 bytes data]
20:03:25.189090 * We are completely uploaded and fine
20:03:25.259254 * Mark bundle as not supporting multiuse
20:03:25.259288 * HTTP 1.0, assume close after body
20:03:25.259307 < HTTP/1.0 200 OK
20:03:25.259322 < Server: BaseHTTP/0.6 Python/3.8.5
20:03:25.259335 < Date: Mon, 01 Mar 2021 18:03:25 GMT
20:03:25.259350 < Content-Type: application/json; charset=UTF-8
20:03:25.259364 < 
20:03:25.259422 { [9051 bytes data]
20:03:25.259531 * Closing connection 0
[
  {
    "label": "art",
    "score": "0.86079"
<...>
```

There's a 1 second delay between:

```
20:03:24.186549 > Expect: 100-continue
20:03:24.186549 > 
```

and:

```
20:03:25.188675 * Done waiting for 100-continue
```

That turns out to be [HTTP/1.1 way](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect) for the client to ask the server preemptively whether it should continue with sending the rest of the data and if it should expect for it to be accepted.

I didn't read much into it, but I think the client (`curl`) assumes that our server supports HTTP/1.1, sends the `Expect:` header and waits for the response, but Python's `BaseHTTPRequestHandler` (which we [use for writing HTTP services](https://docs.python.org/3/library/http.server.html)) doesn't have HTTP/1.1 enabled by default and thus doesn't bother with responding to no `Expect:` headers; `curl` then gives up after one second and the whole connection gets downgraded to good old HTTP/1.0, the penalty being a whole second for each request.

So, the ultimate fix turned out to be:

* Setting `protocol_version = "HTTP/1.1"` static variable in almost every subclass of `BaseHTTPRequestHandler` that we use;
    * One exception is `HashServer` which is our DYI HTTP server that we use for testing our HTTP client and various APIs;
* Adding `Content-Length` in all `BaseHTTPRequestHandler` responses as this is what HTTP/1.1 requires.

This above is just something that I didn't know before so I figured I'll share. Also, this is probably something to keep in mind when writing HTTP services, especially with `BaseHTTPRequestHandler`.

Other than that, I reshuffled some code back and forth. James, could you have a quick look at whether I might have messed something up or you think something could be done in a better way?


jtotoole (approved):

jtotoole:
Oof, sounds like quite the odyssey. This looks like a sensible approach to me, very nicely done!

PR #780, opened 2021-04-01
update stopwords lists

jtotoole:
These updated lists are the result of my collating our existing lists with those from [Stopwords ISO repos](https://github.com/stopwords-iso/), deduplicating, running the new lists through Google Translate in the case of non-English, and finally going through line by line to sanity-check them, erring on the side of removing stop words. 

The only ones with super substantial revisions are English and Portuguese. Adding words from the Stopwords ISO repos generally made the lists longer, and IMO better as those lists incorporated a broader range of non-meaningful terms.


apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -4178,506 +1003,231 @@ there's
...
-trouble
-troubled
-troubles
-truck
-trucks
-true
-truly
-trust
-truth
+trillion
```
dsjen, line 1065: I checked out a "biden" query and trillion stood out to me as a term that was in the old list and not in the new one. @ebndulue - do you have opinions about whether this should be included as a stop word?

jtotoole: @ebndulue here's a link to the new stop words (probably easier to read than the github diff): https://github.com/mediacloud/backend/blob/91cb22c2e44ae6b3323d2459ce48f1296147386f/apps/common/src/python/mediawords/languages/en/en_stop_words.txt

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,1000 @@
...
-investment
-invited
-involve
-involved
-involves
-involving
-inward
-iron
+io
+iq
```
pypt, line 557: It looks like the new list has a full list of country-level TLDs in it. I'd skip adding those as, for example, `.iq` for Iraq could accidentally filter out IQ the intelligence quotient, at least for the first pass. We might find out that we need to re-add them later though.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,1000 @@
...
-positions
-positive
-possessed
-possession
-possibilities
-possibility
 possible
 possibly
-post
-posted
```
pypt, line 3073: We'll probably need to readd stuff like "posted", "published" etc. later on as those are common extraction artifacts ("This blog post was posted on May 1, 2019"). But let's keep it absent for now for the sake of a clean slate.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,1000 @@
...
-search
-searching
-season
-seat
-seated
+sb
+sc
+sd
+se
+sec
```
pypt, line 875: That's the Securities and Exchange Commission! :)

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
pypt: I'd get rid of "trillion" in the new list (and possibly other numbers) but IMHO something to keep in mind is that stopwords are very context-specific, and while "trillion" is not significant in many contexts, it would be in the context of the [Trillion Tree Campaign](https://en.wikipedia.org/wiki/Trillion_Tree_Campaign) coverage, so I'd rather keep the backend-filtered list very small (skip the "is", "not" etc.) and do most of the filtering on the frontend using the subject matter basis (perhaps a saved search could have a custom stopword list functionality?)

apps/common/src/python/mediawords/languages/lt/lt_stop_words.txt
```diff
@@ -1,19 +1,78 @@
...
+anuosiuose
+anuosius
 apie
+aplink
 ar
 arba
+argi
+arti
+ástriþai
+aukðèiau
```
pypt, line 65: Here and elsewhere, you didn't get the character set right :) The list that you've looked at was probably in ISO-8859-13, and this list needs to be in UTF-8. In other words, `aukðèiau` should have been `aukščiau`.

apps/common/src/python/mediawords/languages/lt/lt_stop_words.txt
```diff
@@ -78,56 +261,280 @@ metais
...
+ðiuodu
+ðiuoju
+ðiuosiuose
+ðiuosius
+ðtai
+þemiau
 į
 šalia
-šalies
+čia
```
pypt, line 538: `čia` is fine though, so probably you took it from some other list.

apps/common/src/python/mediawords/languages/ru/ru_stop_words.txt
```diff
@@ -16,894 +21,691 @@ rnd
...
+с
+т
+у
+х
+я
+с кем
 АО
+Во
+Вы
+Да
```
pypt, line 61: Could you sort this list and others, also remove duplicates if there are any? At least Sublime Text has multilanguage-aware sorting functionality (Edit -> Sort Lines and Edit -> Permute Lines -> Unique might come in handy).

apps/common/src/python/mediawords/languages/tr/tr_stop_words.txt
```diff
@@ -1,101 +1,238 @@
...
 # This is a stop word list for the Turkish language.
+# (Lightly edited to remove words in the original lists that are actually meaningful)
 #
 # Sources:
 #     http://nlp.ceng.fatih.edu.tr/blog/?p=101
 #     http://www.ranks.nl/stopwords/turkish.html
+#     https://github.com/stopwords-iso/stopwords-tr/blob/master/stopwords-tr.txt
 #
 
+INSERmi
```
pypt, line 10: Insermi looks like a brandname of sorts to me.

pypt (changes requested):
Some minor changes please, otherwise looks okay to me.

We'll 100% need to amend it afterwards, e.g. readd things like `blockquote` (it's a common extraction artifact from the tag `<blockquote>`) or `published` (mentioned in a lot of articles near the publication date) but I'd be a proponent of starting that work from a clean slate such as this.

Just to reiterate - having less here is better than having more as, for example, Trillion Tree Campaign researchers would be unhappy about us filtering out the term "trillion" system-wide, so this stopwords list has to be a compromise that makes all parties dissatisfied (to paraphrase Larry David), and one way to address the unhapiness would be per-search stopwords (stored within a saved search?)

pypt:
@jtotoole could you have a quick look at the failing tests too? I think most (all?) of the failing tests are the language ones (`test_XX.py`).

apps/common/tests/python/mediawords/languages/test_lt.py
```diff
@@ -17,7 +17,7 @@ def test_sample_sentence(self):
 
     def test_stop_words_map(self):
         stop_words = self.__tokenizer.stop_words_map()
-        assert "buvo" in stop_words
```
pypt, line 20: `buvo` literally means `was` in Lithuanian :) Weird that the "ISO" stopwords list didn't include a term like that.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-cellar
-cells
-cent
-center
-centers
-central
-cents
-centuries
-century
+cc
```
pypt, line 253: CC is a TLD, and it could also mean a credit card.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
 buy
-buying
 by
 c
 c'mon
 c's
-cafe
-calculated
-calendar
+ca
```
pypt, line 241: CA is a TLD as well.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-arrived
-art
-artery
-article
-articles
-artist
-artistic
-artists
-arts
+arpa
```
pypt, line 201: `arpa` is also a TLD and a name of an important project.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-fence
-festival
-few
-fewer
-fiber
-fibers
-fiction
-field
-fields
+fi
```
pypt, line 336: Also a TLD.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-filled
-filling
-film
-films
-final
-finally
-finance
-financial
-financing
+fify
```
pypt, line 340: Does this mean *Fixed it for you*?

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-followed
-following
-follows
-font
-food
-foods
-fool
-foot
-football
+fix
```
pypt, line 344: I'd skip `fix` as well, at least for the initial pass of stopwords. Someone might be researching into someone's attempts to "fix" the economy for example. If "fix" is a stopword then a lot of other words from the removed list should belong in the list of stopwords as well.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
 hr
-href
+ht
+htm
 html
 http
-huge
-human
-humanity
+hu
```
pypt, line 437: TLD.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-ideas
-identical
-identification
-identified
-identify
-identity
-ideological
+i.e
+i.e.
+id
```
pypt, line 446: TLD.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-impression
-impressions
-impressive
-improve
-improved
-improvement
-improvements
-impulse
+ii
+il
```
pypt, line 450: TLD.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
+ill
+im
```
pypt, line 452: TLD (I think).

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-internal
-international
-interpretation
-interpreted
-intervals
-intervention
-interview
-interviews
-intimate
+int
```
pypt, line 463: Maybe remove `int` here too?

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-miles
-military
-milk
-milligrams
-million
-millions
-mind
-minds
+mightnt
+mil
```
pypt, line 539: TLD.

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-multiply
-municipal
-murder
-muscle
-muscles
-music
-musical
-musician
-musicians
+mug
```
pypt, line 552: No research into coffee cups :(

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -1,4175 +1,798 @@
...
-orders
-organ
-organic
-organization
-organizations
-organized
-origin
-original
-originally
+org
```
pypt, line 618: TLD (again, we might have to remove it afterwards but let's just skip it on the first pass to see what we get with a very non-restrictive list of stopwords).

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -4178,506 +801,195 @@ there's
...
-upper
-upstairs
-upward
-upwards
-urban
-urge
-urged
-urgent
-url
+ups
```
pypt, line 873: Could you resurrect this brandname?

apps/common/src/python/mediawords/languages/en/en_stop_words.txt
```diff
@@ -4178,506 +801,195 @@ there's
...
 us
```
pypt, line 874: Idea for a fun flamewar in the future: `us` is definitely a stopword but `US` is an abbreviation of a country which could be considered a stopword in the US media but not so much in other English-speaking countries. So perhaps those stopwords could be made case-sensitive at some point?

apps/common/src/python/mediawords/languages/lt/lt_stop_words.txt
```diff
@@ -3,131 +3,481 @@
 #
 # Sources:
 #     http://www.filewatcher.com/p/punbb-1.2.16.tbz.620109/www/punbb/lang/Lithuanian/stopwords.txt.html
+#     https://github.com/stopwords-iso/stopwords-lt/blob/master/stopwords-lt.txt
 #     auto-generated sources
 #
 
-a
+abi
```
pypt, line 10: The Lithuanian stopword list is not particularly important for us as no Lithuanians do research with Media Cloud (some did try it in the past though, and one friend of mine even defended her Master's using the tool), but in this particular case both the old stopword list and the new one should be joined together as both include very valid, legit, one-size-fits-all stopwords.

apps/common/src/python/mediawords/languages/ru/ru_stop_words.txt
pypt: Somehow the thing still didn't get sorted :( Could you resort other lists too?

pypt (changes requested):
Some more cleanup please, esp. TLDs.

apps/common/src/python/mediawords/languages/lt/lt_stop_words.txt
pypt: Very cool, thank you very much!

pypt (approved):
Good to go, thank you for your work James!

pypt:
Merged into `master` and `release` for all of it to get rebuilt, not deployed yet.

PR #781, opened 2021-04-01
Add API hack to show word clouds filtered with old and new stopword lists

pypt:
I've duplicated the current stopword lists as such:

* `ca/ca_stop_words.txt` - the updated list of stopwords once @jtotoole finalizes the list
* `ca/ca_stop_words_old.txt` - the old, overly aggressive list of
stopwords
* ...

then did some ugly copy-pasting of some API code for the relevant
endpoints to support the `old_stopwords` parameter. Once set to
something like `1`, the API call will do the word cloud filtering using
the "old" stopword list instead of the "new" one. Example:

* `stories/list?q=...` - uses the new, non-aggressive stopword list;
* `stories/list?q=...&old_stopwords=1` - uses the old, overly aggressive
stopword list.

The idea is that once @jtotoole comes up with his final list of "new"
stopwords, we can merge his list on top of this PR, deploy, and that way
we'll get a rather easy way to compare stopwords.

API calls that use stopword filtering to create word clouds and thus
support the `old_stopwords` parameter:

* `wc/list`
* `stories/word_matrix`
* any call that lists stories (e.g. `stories/list`) *and* passes the `show_wc` parameter (so @dsjen, you can find all uses of the call by grepping for `show_wc`)

https://github.com/mediacloud/api-client hasn't been updated
accordingly so you might want to do that too. Also, I have no idea how
the caching layer might treat this as I'm not familiar with it.

@dsjen, is this something that you could use on your end?

If so, @jtotoole, could you rebase your stopword branch on top of this
one?


dsjen:
thanks, Linas! This makes comparison so much simpler and direct! I can make the changes to the client.

jtotoole:
very clever! thanks for the quick work 🙂

jtotoole (approved):

pypt:
Merged, will let you know when it gets deployed.

PR #782, opened 2021-04-13
Resurrect `mail` utility for munin-cron, generate `alerts.conf` dynamically

pypt:
Munin is our monitoring tool of choice for all app-related matters. It doesn't (and is not supposed to) do any kind of systems monitoring as Munin gets run as a Docker service, so if the `mediacloud` Docker stack is down, that makes Munin down as well. Instead, it tracks various app-related metrics such as:

* How many stories got imported into Solr during last hour: https://github.com/mediacloud/backend/blob/63545b1f3cd02f2e170801cd90842e5a8f7751a0/apps/munin-node/plugins/mc_solr_stories_last_hour
* How many downloads didn't yet get extracted: https://github.com/mediacloud/backend/blob/63545b1f3cd02f2e170801cd90842e5a8f7751a0/apps/munin-node/plugins/mc_downloads_unextracted
* How many media sources do we have: https://github.com/mediacloud/backend/blob/63545b1f3cd02f2e170801cd90842e5a8f7751a0/apps/munin-node/plugins/mc_media_count
* ...and others: https://github.com/mediacloud/backend/tree/63545b1f3cd02f2e170801cd90842e5a8f7751a0/apps/munin-node/plugins

We chose Munin because:

* It's rather simple ("plugins" are simple scripts that output a bunch of numbers in a specific format so you can write them in any language);
* It's able to send alert emails when things look off (unlike some other tools);
* It's even able to send alert emails when it's unable to figure out if things are off (unlike most other tools).

`munin-cron` service polls `munin-node` service periodically (every 5 minutes) for stats, writes them down, tests whether those stats are out of bounds (as defined in the plugin's source code) and if they look off, sends us an email.

Munin is super old and so it doesn't know how to connect to the SMTP server (`mail-postfix-server`) directly. Instead, it calls Linux's `mail` utility (https://linux.die.net/man/1/mail; somewhat similar to `sendmail`) to submit the alert email to be sent.

Weird Linux (or UNIX?) history: just like `sendmail`, `mail` is not so much of a single utility that you can install but rather a "protocol" that can be implemented by various vendors which provide utilities compatible with `mail` and `sendmail`. The idea is that one can choose a provider (a specific implementation of `mail` or `sendmail`) to do their `mail`ing or `sendmail`ing, and then the tools that call `mail` or `sendmail` will be able to send the email while not caring how exactly does it get sent.

We were using S-nail (https://wiki.archlinux.org/index.php/S-nail) for our `mail` needs, and msmtp (https://wiki.archlinux.org/index.php/msmtp) to do the `sendmail`ing. `mail` (implemented via S-nail) was pre-configured to call `sendmail` utility to submit outgoing email to it, and `sendmail` (implemented via msmtp) then was to connect to `mail-postfix-server` and post the outgoing message. The end result in all cases was the same - no matter how you choose to send email from any container (`mail`, `sendmail`, or direct connection to the SMTP server), the message would still go out.

So, but it then turns out that Debian (probably) people changed up how S-nail (`s-nail` APT package) gets installed, and it stopped providing `mail` utility, the end result being that `munin-cron` service was silently failing at sending us alert emails about problems. Not good!

So, I changed it up as follows:

1) Instead of S-nail, we're now installing `bsd-mailx` package which also provides the `mail` utility which is also pre-configured to submit emails to `sendmail`;
2) For extra paranoia, we're now verifying whether `sendmail` is symlinked to `msmtp` and `mail` is symlinked to `/usr/bin/bsd-mailx`

Plus, the email addresses to send alerts to were hardcoded so I've made them configurable by auto-generating `alerts.conf` file on container startup. We'll point it to Google Groups after deployment so that all of us could get alerts about things going south.

James, could you have a look?


jtotoole:
makes sense and looks good! clever trick with the `symlink` check

jtotoole (approved):

PR #783, opened 2021-04-14
Spruce up English stop words

jtotoole:
I went into this PR with the plan to add a list and fill in some gaps, but ended up doing more removing of potentially meaningful words than adding missing ones. Feel free to just close this out if you don't think it's useful.


pypt:
Thank you for the initiative but I think the sprucing up went the wrong way - we need more stopwords here, not less. For example, IMHO the term `possibly` is not potentially useful to anyone in any context whatsoever.

In any case, this is a flamewar that we might have at some point in the future, but for now (as agreed in the team meeting) let's do nothing regarding stopwords until 1) we need the word cloud to get updated; 2) someone complains.

PR #787, opened 2021-05-06
Raise NotImplementedError exception, not NotImplemented which is a null value

pypt:
Awkward typo of mine.

* https://docs.python.org/3/library/constants.html#NotImplemented
* https://docs.python.org/3/library/exceptions.html#NotImplementedError


jtotoole:
Implement away! :rocket:

jtotoole (approved):

PR #788, opened 2021-05-10
Munin fixes

pypt:
* Remove "mc_solr_sentences_last_day" plugin. We used to index stories as individual sentences in Solr. That's no longer the case for quite some time, so this plugin is no longer useful (nor does it work really).
* Fix expected string in "mc_websites_up" plugin


pypt:
This is to fix some of the alerts that we've got just now. Makes me wonder why did we get those alerts today and not earlier though :)

jtotoole:
LGTM 👍

jtotoole (approved):

PR #791, opened 2021-05-29
Use a window function instead of a recursive CTE for adding downloads to crawler-fetcher queue, add download IDs in chunks

pypt:
On Friday evening (of course!), crawler has decided that it doesn't feel like fetching feeds / stories anymore.

Short intro into how everything works:

* `crawler-provider` decides which feeds / stories / downloads are due to be fetched and adds them to `queued_downloads` table;
* `crawler-fetcher` polls the `queued_downloads` table to find the downloads to fetch.

`crawler-provider` has to identify which downloads from `downloads_pending` table it has to get enqueued for fetching, but it has to add only one download per host (e.g. one download for `nytimes.com`) to the `queued_downloads` table so that the fetchers don't end up fetching 1000 URLs from a single host in parallel. It used to do this task with the following plpgsql function:

```sql
-- efficiently query downloads_pending for the latest downloads_id per host.  postgres is not able to do this through
-- its normal query planning (it just does an index scan of the whole index).  this turns a query that 
-- takes ~22 seconds for a 100 million row table into one that takes ~0.25 seconds
create or replace function get_downloads_for_queue() returns table(downloads_id bigint) as $$
declare
    pending_host record;
begin
    -- quick temp copy without dead row issues for querying in loop body below
    create temporary table qd on commit drop as select * from queued_downloads;

    create temporary table pending_downloads (downloads_id bigint) on commit drop;
    for pending_host in
            WITH RECURSIVE t AS (
               (SELECT host FROM downloads_pending ORDER BY host LIMIT 1)
               UNION ALL
               SELECT (SELECT host FROM downloads_pending WHERE host > t.host ORDER BY host LIMIT 1)
               FROM t
               WHERE t.host IS NOT NULL
               )
            SELECT host FROM t WHERE host IS NOT NULL
        loop
            insert into pending_downloads
                select dp.downloads_id
                    from downloads_pending dp
                        left join qd on ( dp.downloads_id = qd.downloads_id )
                    where 
                        host = pending_host.host and
                        qd.downloads_id is null
                    order by priority, downloads_id desc nulls last
                    limit 1;
        end loop;

    return query select pd.downloads_id from pending_downloads pd;
 end;

$$ language plpgsql;
```

The function:

1. Read unique (distinct) `host`s from `downloads_pending` one by one;
2. Fetched the most prioritized download from `downloads_pending` individually.

I think it was implemented like that due to some sort of a performance reason (sometimes it's hard to make PostgreSQL use its indexes in the most efficient manner). Whatever the reasoning, the function took 2 hours to complete on every run which made it infeasible, so I've simplified things by making `crawler-provider` run a [window function](https://stackoverflow.com/a/3800572).

Also, upon collecting download IDs to fetch (after two hours), `crawler-fetcher` used to aggregate them all into a single string to later insert into `queued_downloads`:

```sql
values = ','.join(["(%d)" % i for i in downloads_ids])
db.query(
    "insert into queued_downloads(downloads_id) values %s on conflict (downloads_id) do nothing" %
    values)
```

Two problems here:

1. Our database handler accepts Python `list`s as parameters which it then turns into PostgreSQL `ARRAY[]`s so we can capitalize on that fact and make things simpler;
2. More importantly, if you so happen to have a lot (tens, hundreds of thousands) of download IDs to be inserted, the code above will generate a single massive SQL query which will take ages to parse, run, and most probably will just fail.

So, now the list of download IDs get split into chunks of a thousand each and inserted separately.

@jtotoole could you please have a look at this one? It's already deployed (to not lose stories over the weekend) but maybe I've made errors in my plpgsql -> window function rewrite. This one very much is in need of a second pair of eyes because it's the critical path of our codebase.

### DIFF
--- a/apps/crawler-provider/src/python/crawler_provider/__init__.py
+++ b/apps/crawler-provider/src/python/crawler_provider/__init__.py
@@ -15,12 +15,13 @@ once per second.
 
 The provider works as a daemon, periodically checking the size queued_downloads and only adding
 new jobs to the queue if there are more than MAX_QUEUE_SIZE jobs in the table.  This allows us to implement
-throttline by keeping the crawler jobs queue relatively small, thus limiting the number of requests for each
+throttling by keeping the crawler jobs queue relatively small, thus limiting the number of requests for each
 host over a period of several minutes, while allowing the crawler_fetcher jobs to acts as simple stupid
 worker jobs that just do a quick query of queued_downloads to grab the oldest queued download.
 """
 
 import time
+from typing import List
 
 from mediawords.db import DatabaseHandler
 from mediawords.util.log import create_logger
@@ -97,7 +98,7 @@ def _add_stale_feeds(db: DatabaseHandler) -> None:
             -- Feed was downloaded more than stale_feed_interval seconds ago
             OR (last_attempted_download_time < (NOW() - (%(a)s || ' seconds')::interval))
 
-            -- (Probably) if a new story comes in every "n" seconds, refetch feed every "n" + 5 minutes
+            -- (Probably) if a new story comes in every "n" seconds, re-fetch feed every "n" + 5 minutes
             OR (
                 (NOW() > last_attempted_download_time +
                         (last_attempted_download_time - last_new_story_time) + interval '5 minutes')
--- a/apps/crawler-provider/src/python/crawler_provider/__init__.py
+++ b/apps/crawler-provider/src/python/crawler_provider/__init__.py
@@ -196,7 +196,7 @@ def provide_download_ids(db: DatabaseHandler) -> List[int]:
     
     """).flat()
 
-    log.info(f"Provide downloads host downloads: {len(downloads_ids)}")
+    log.info(f"Providing {len(downloads_ids)} per-host download IDs")
 
     return downloads_ids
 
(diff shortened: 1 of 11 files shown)


apps/crawler-provider/src/python/crawler_provider/__init__.py
```diff
@@ -158,17 +164,49 @@ def provide_download_ids(db: DatabaseHandler) -> None:
...
+        )
+        
+        SELECT downloads_id
+        FROM pending_downloads_per_host
+        WHERE rank = 1
+    
+    """).flat()
 
-    log.info("provide downloads host downloads: %d" % len(downloads_ids))
+    log.info(f"Provide downloads host downloads: {len(downloads_ids)}")
```
jtotoole, line 199: This logging is confusing without context—does this essentially mean, "Pass the ID of the next story to be downloaded for each host to the downloads queue"?

jtotoole:
Looks good to me! The window function makes a lot more sense than the custom plpgsql, and it's easy to see how things could go wrong with the "insert one massive concatenated string" approach. Added points for tackling this on a weekend!

jtotoole (approved):

apps/crawler-provider/src/python/crawler_provider/__init__.py
pypt: Thanks, good point, improved the log message slightly. The query comes up with a list of unique per-host download IDs to be fetched, nothing to do with stories yet. Let me know if it's not that much better.

PR #792, opened 2021-06-04
Upgrade PostgreSQL from 11 to 13, add upgrade helper script

pypt:
We're running a pre-COVID version 11 of PostgreSQL which is not that old but we could really use some features introduced in 12 and 13. My favorites:

[**12**](https://www.postgresql.org/about/news/postgresql-12-released-1976/):

* _B-tree Indexes, the standard type of indexing in PostgreSQL, have been optimized in PostgreSQL 12 to better handle workloads where the indexes are frequently modified. Using a fair use implementation of the TPC-C benchmark, PostgreSQL 12 demonstrated on average a 40% reduction in space utilization and an overall gain in query performance._
    * I think we would have to recreate those indexes to get the space utilization benefit though.
* _Queries on partitioned tables have also seen demonstrable improvements, particularly for tables with thousands of partitions that only need to retrieve data from a limited subset. PostgreSQL 12 also improves the performance of adding data to partitioned tables with INSERT and COPY, and includes the ability to attach a new partition to a table without blocking queries._
* _There are additional enhancements to indexing in PostgreSQL 12 that affect overall performance, including <...> CREATE STATISTICS now supporting most-common value (MCV) statistics to help generate better query plans when using columns that are nonuniformly distributed._
    * If I read this correctly, with `CREATE STATISTICS` one can now micromanage if and how the database uses indexes on cases when a table's row distribution is not uniform, e.g. if you have a lot of 1's somewhere in the middle of a bunch of 0's.
* _Common table expressions, also known as WITH queries, can now be automatically inlined by PostgreSQL 12, which in turn can help increase the performance of many existing queries. In this release, a WITH query can be inlined if it is not recursive, does not have any side-effects, and is only referenced once in a later part of a query._
    * CTEs are the ones with `WITH`, e.g. the one from https://github.com/mediacloud/backend/pull/791. Now one can choose whether PostgreSQL has to run the expression within `WITH`, materialize it, and lastly pass it to the next expression, or should it be allowed to be smart and work out what it things is the best way to run everything in one go. More often than not, we need PostgreSQL to be *less* smart about things and run `WITH` expressions isolated from each other.
* _PostgreSQL 12 introduces "generated columns." Defined in the SQL standard, this type of column computes its value from the contents of other columns in the same table. In this version, PostgreSQL supports "stored generated columns," where the computed value is stored on the disk._
    * We have a few of those anywhere that are currently generated in the application.
* _PostgreSQL 12 introduces the ability to rebuild indexes without blocking writes to an index via the REINDEX CONCURRENTLY command, allowing users to avoid downtime scenarios for lengthy index rebuilds._
    * One less command to type (`REINDEX CONCURRENTLY` instead of `CREATE INDEX CONCURRENTLY` + `ALTER INDEX ... RENAME TO`)

[**13**](https://www.postgresql.org/about/news/postgresql-13-released-2077/):

* _Building on work from the previous PostgreSQL release, PostgreSQL 13 can efficiently handle duplicate data in B-tree indexes, the standard database index. This lowers the overall space usage that B-tree indexes require while improving overall query performance._
* _PostgreSQL 13 introduces incremental sorting, where sorted data from an earlier step in a query can accelerate sorting at a later step. Additionally, PostgreSQL can now use the extended statistics system (accessed via CREATE STATISTICS) to create improved plans for queries with OR clauses and IN/ANY lookups over lists._
* _Queries with partitioned tables have received a performance boost, as there are now more cases where partitions can be pruned and where partitions can be directly joined._
* _PostgreSQL 13 continues to improve the vacuuming system with the introduction of parallelized vacuum for indexes. In addition to the vacuuming performance benefits it offers, the use of this new feature can be tuned to specific workloads as administrators can select the number of parallel workers to run. In addition to these performance benefits, data inserts can now trigger the autovacuum process._
    * [ ] To-do here: figure out how one configures that vacuuming parallelism.
* _PostgreSQL 13 also adds more ways an administrator can monitor database activity, including referencing WAL usage statistics from EXPLAIN, the progress of streaming base backups, and the progress of the ANALYZE commands._
    * Progress on `VACUUM` is amazing as those take ages. Plus, at some point they've added progress reporting on `CREATE INDEX` too.

Lastly, Citus 10, a great improvement of a database sharding solution that I've been eyeing, [requires PostgreSQL 12](https://www.citusdata.com/blog/2021/03/05/citus-10-release-open-source-rebalancer-and-columnar-for-postgres/).

In addition to the rather uneventful upgrade, I've added a `postgresql-upgrade` app with a Python script that should make such upgrades easier, plus `doc/postgresql_upgrade.markdown` describing how one should do such an upgrade. @jtotoole could you have a look at it to see whether you find this doc readable?


pypt:
Upgrade done, log attached.

[postgresql_11_to_13_upgrade.log](https://github.com/mediacloud/backend/files/6608775/postgresql_11_to_13_upgrade.log)

jtotoole:
`README` looks good! Some distant future us will be thankful.

PR #793, opened 2021-06-23
[RFR] remove crimson hexagon references

jtotoole:
Decided to make a fresh PR rather than fight my way out of merge commit hell on https://github.com/mediacloud/backend/pull/785. 

Does adding `select insert_platform_source_pair( 'twitter', 'brandwatch' );` [here](https://github.com/mediacloud/backend/blob/master/apps/postgresql-server/schema/mediawords.sql#L1824) in `mediawords.sql` entail a migration, and should I also delete `select insert_platform_source_pair( 'twitter', 'crimson_hexagon' );`?


jtotoole:
@pypt now we're in business (RFR = Ready For Review).

apps/crawler-fetcher/tests/python/test_podcast_resolve_url.py
```diff
@@ -34,21 +34,21 @@ def test_get_feed_url_from_google_podcasts_url():
     assert _get_feed_url_from_google_podcasts_url('http://www.example.com/') == 'http://www.example.com/'
     assert _get_feed_url_from_google_podcasts_url('totally not an URL') == 'totally not an URL'
 
-    npr_feed_url = 'https://feeds.npr.org/381444908/podcast.xml'
```
jtotoole, line 37: this feed url seems to oscillate at times for no reason, so hopefully the FT one will be more stable

pypt:
> Does adding `select insert_platform_source_pair( 'twitter', 'brandwatch' );` [here](https://github.com/mediacloud/backend/blob/master/apps/postgresql-server/schema/mediawords.sql#L1824) in `mediawords.sql` entail a migration, and should I also delete `select insert_platform_source_pair( 'twitter', 'crimson_hexagon' );`?

I think it does - that function preloads the table with some platform-source pairs, so you might want to `DELETE` from the table that it preloads into to get rid of that deprecated platform-source pair.

pypt:
> I think it does - that function preloads the table with some platform-source pairs, so you might want to DELETE from the table that it preloads into to get rid of that deprecated platform-source pair.

...and all of the rows in other tables with reference that particular row in `topic_sources`.

jtotoole:
> ...so you might want to DELETE from the table that it preloads into to get rid of that deprecated platform-source pair

Would it be okay to do that in a separate PR once the pgmigrate one gets merged?

pypt:
Very cool, thank you! Down with CH!

pypt:
> Would it be okay to do that in a separate PR once the pgmigrate one gets merged?

Sure, moved this to https://github.com/mediacloud/backend/issues/798,

PR #797, opened 2021-07-10
Move to docker hub from gcr

jtotoole:
Related PR: https://github.com/mediacloud/systems/pull/229
Original issue: https://github.com/mediacloud/backend/issues/794

@pypt Do you think we need to pass the new Docker Hub credentials as args in the dev scripts, along the lines of what was reversed in [this PR](https://github.com/mediacloud/backend/commit/80890c5c1ff8bb7bae4769fc6d1f0259f0fccf64) (e.g. [here](https://github.com/mediacloud/backend/commit/80890c5c1ff8bb7bae4769fc6d1f0259f0fccf64#diff-59e3d77b55c5d33e7116af25e97746b21a6af9d1c5d42b74bff875c99d43fb06L23))? Even without such arguments added to those dev scripts, the images can be built and pushed correctly at present (new Docker Hub repos [here](https://hub.docker.com/orgs/mc2021/repositories)).


pypt:
No, why pass those credentials? One is supposed to "docker login" using their own Docker Hub user on their own laptop to be able to push something. If someone quits Media Cloud, we would then have to change those shared credentials on all developer computers instead of simply removing a specific user from Docker Hub organisation.

jtotoole:
Cool that was my assumption, just wanted to make sure. The only test fails here are related to Crimson Hexagon, so I think this is good to go assuming it right solid to you.

.github/workflows/build.yml
```diff
@@ -50,10 +50,8 @@ jobs:
           docker version
           docker-compose version
 
-      - name: Log in to container registry
-        run: echo "${DOCKER_GCR_SERVICE_ACCOUNT}" | docker login -u _json_key --password-stdin https://gcr.io
-        env:
-          DOCKER_GCR_SERVICE_ACCOUNT: ${{ secrets.DOCKER_GCR_SERVICE_ACCOUNT }}
+      - name: Log into Docker Hub
+        run: echo "${{ secrets.DOCKERHUB_PASSWORD }}" | docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
```
pypt, line 54: Are you using the `dockermediacloudtravis` user here and a personal access token instead of a password?

apps/docker-compose.dist.yml
```diff
@@ -181,7 +181,7 @@ x-brandwatch-api-configuration: &brandwatch-api-configuration
 # "deploy/replicas", we define every shard as its own independent service.
 #
 x-solr-shard_base:   &solr-shard_base
```
pypt, line 183: I remembered why using Docker Hub with with Docker Swarm was painful - even if one does `docker login` to log in to the authenticated (and $50 lighter) account belonging to an organization, Docker daemon [doesn't bother passing those credentials to the Hub](https://docs.docker.com/docker-hub/download-rate-limit/#docker-swarm) unless [one passes the `--with-registry-auth` argument with `docker stack deploy`](https://docs.docker.com/engine/reference/commandline/stack_deploy/).

So, could you add this argument to the "deploy this thing by running this" comment at the top of this file and elsewhere (if appropriate)?

doc/dev_scripts.markdown
```diff
@@ -37,7 +37,7 @@ Also, almost every script (`print_docker_run_in_stack.py` being a notable except
 
 ## `pull.py` - pull images
```
pypt, line 38: And remove `doc/docker_gcr.markdown` if you can.

pypt (changes requested):
Just a few minor things.

.github/workflows/build.yml
jtotoole: No, I added new GitHub secrets for the `mcsystems` user in the `mc2021` Docker Hub org

apps/docker-compose.dist.yml
jtotoole: well-spotted!

.github/workflows/build.yml
pypt: Could you either reuse `dockermediacloudtravis` user for pulling / pushing those images, or (perhaps better) add separate Docker Hub users for:

1) CI (currently GitHub Actions) pulling / pushing the images;
2) Production servers pulling the images.

All of these users should have 2FA enabled, use personal access tokens instead of their passwords, know nothing about billing, and have only appropriate permissions to pull from / push to Docker Hub (e.g. the user doing the pulling needs only read-only permissions - I don't know if that's doable on Docker Hub); we'd then be using `mcsystems` only for billing and administrative tasks.

I've made the mistake of calling `dockermediacloudtravis` user after the specific CI that we were using at the time (Travis CI) so don't mention GitHub Actions or any specific "brandnames" in the usernames :)

PR #800, opened 2021-07-27
drop cliff_ + nytlabels_annotations tables via pgmigrate

jtotoole:
relates to https://github.com/mediacloud/backend/issues/743


pypt:
Fixes #743.

PR #801, opened 2021-07-27
Fix typo in path to pgmigrate migrations

pypt:
Migrations don't get applied on first run of the container so this wasn't caught by the CI.

A simple way to make sure that migrations work is to start some sort of a test Compose layout, e.g.:

```bash
./dev/run.py common bash
```

and then restart the postgresql-server container:

```bash
docker restart $(docker ps | grep postgresql-server | awk '{ print $1 }')
```


jtotoole:
ugh, search/replace fail. thanks for catching! merge away when you're ready

jtotoole (approved):

PR #803, opened 2021-08-06
add doc on dumping postgresql data locally

jtotoole:


doc/docker_cheat_sheet.markdown
```diff
@@ -152,3 +153,25 @@ To access PostgreSQL directly, you can either run `psql` in a `postgresql-server
...
      mediacloud=# 
      ```
+
+* To **dump PostgreSQL tables to your host machine**:
+
+   1. `ssh` to the server running PostgreSQL and find its container ID, as described above.
+   
+   2. Choose the table you want and dump it as a `.sql` file with the following syntax:
+
+      ```docker exec 123somecontainer456 pg_dump --table=media > media.sql`
```
jtotoole, line 163: I found that this only worked when I omitted the `-it` flags

PR #806, opened 2021-08-23
Use testmediacloud.tk instead of testmediacloud.ml

pypt:
`testmediacloud` is our test domain that we use for testing DNS changes,
e.g. IPv6 addressing to LXC containers, email sending / receiving etc.

Domain is registered on Freenom.com. The `.ml` version expired, and they
wanted 8 EUR for a renewal, so I went with `.tk` which was free.

[ci skip]


jtotoole:
:shipit:

jtotoole (approved):

Issue #812, opened 2021-10-08
Back up PostgreSQL databases


pypt:
Given that disks on which we run our services are pre-shot or at least very old, and the disk health review + replacement is still ongoing, we need to improve PostgreSQL backups by implementing the [3-2-1 strategy](https://www.backblaze.com/blog/the-3-2-1-backup-strategy/).

Current state of affairs is as follows:

1. First copy of the PostgreSQL dataset: production dataset on woodward. Runs on disks that are still alive but are super old.
2. Second copy of the PostgreSQL dataset: in `-systems` repo, we have a `zfs-send.py` script which used to run once a day and would [copy a ZFS snapshot](https://docs.oracle.com/cd/E18752_01/html/819-5461/gbchx.html) of PostgreSQL data to another server, sinclair. However, we suspended this script due to Holyoke's annual shutdown and haven't resumed its operation since. Plus, sinclair, the server that is being backed up to, has even older disks, at least one of which is shot already, so it's like we 
3. Third copy of the PostgreSQL dataset: largely nonexistent. There's a copy of the dataset from 2019 somewhere, and I have recently manually copied the full dataset to S3, but our data is not being periodically copied anywhere, so if (when) those disks decide to move into eternity (pretty soon now), we'll be able to restore only the dataset that's a few months old.

Therefore, a to-do:

* [ ] Make `zfs-send.py` periodically copy PostgreSQL dataset from woodward to sinclair again. Sinclair's disks are unreliable, but at least it's something.
* [ ] Deploy WAL-G, do a full backup, do a few incremental backups, measure how much time + space do they take.
    * [ ] System load
    * [ ] Time to do full backup
    * [ ] Time to do incremental backup
    * [ ] Requests needed to periodically upload WALs to S3
* [ ] Increase WAL size so that uploads to S3 don't happen every 5 seconds
* [ ] Try + document restoring a WAL-G backup, measure time it would take it for us to do it.
* [ ] Clean up stuff on test S3 buckets.
* [ ] Make WAL-G do a periodic backup every day or so.

pypt:
Started test full backup to S3.

pypt:
Initial backup to B2 exited because the PostgreSQL container went away (got killed by someone / something):

```
<...>
INFO: 2021/11/02 06:54:03.685946 Finished writing part 24382.
INFO: 2021/11/02 06:54:03.686024 Starting part 24383 ...
[...]

PR #813, opened 2021-10-08
Install + set up PgBackRest in postgresql-base

pypt:
Install and configure [PgBackRest](https://pgbackrest.org/) to `postgresql-base` so that images that derive from it - `postgresql-server` and `temporal-postgresql` - could use it.

The only supported repository for backups in this PR is Amazon S3 and S3-compatible object stores. Repository's credentials are configured through (optional) environment variables.

There's a terse doc on how to do full / incremental backups too.

As per #812, I intend to deploy this, do a full backup, see how long does it take + how much space does it occupy, and then come up with a backup strategy.

Unrelated, but in:

https://github.com/mediacloud/backend/compare/pgbackrest?expand=1#diff-2c08109dfb859a1eab9191db0d9aab0b4b330e09c5666f15e5654ca0be3b6933R1930-R1936

`/dev/shm`'s size for `temporal-postgresql` gets increased too. PostgreSQL used to rely on `/dev/shm` heavily, I'm not sure if that's the case anymore, but it doesn't hurt to hike it up a bit.

References #812.

### DIFF
--- a/doc/postgresql_pgbackrest.markdown
+++ b/doc/postgresql_pgbackrest.markdown
@@ -3,6 +3,8 @@
 
 ## Usage
 
+A stanza is the [configuration for a PostgreSQL database cluster that defines where it is located, how it will be backed up, archiving options, etc.](https://pgbackrest.org/user-guide.html)
+
 Create stanza:
 
 ```bash


jtotoole:
This looks really good! Questions just for my understanding:

- From where does one configure backup frequency?
- Is there a way to conditionally increase `MC_PGBACKREST_RETENTION_FULL` if the latest backup has failed? Is that something we'd want to do?

pypt:
> From where does one configure backup frequency?

At the moment backups happen when we run `pgbackrest backup` manually. Later when we know long do they take and how much disk space does a typical backup occupy, we'll add them as a Cron job somewhere.

> Is there a way to conditionally increase MC_PGBACKREST_RETENTION_FULL if the latest backup has failed? Is that something we'd want to do?

Could you clarify this question?

PR #831, opened 2022-03-06
Make no_dedup_sentences the default for extract-and-vector

philbudne:
can be overridden via MC_NO_DEDUP_SENTENCES env var

apps/common/src/python/mediawords/util/config/__init__.py: add env_bool function
apps/extract-and-vector/bin/extract_and_vector_worker.py: honor MC_NO_DEDUP_SENTENCES


philbudne:
NOTE!! This change requires building both gcr.io/mcback/common and gcr.io/mcback/extract-and-vector images!!!

PR #842, opened 2022-11-07
Fix solr collections

thepsalmist:
Fix `java.lang.ClassCastException` bug by querying mediacloud32 and mediacloud64 solr collections separately, then combine responses.

### DIFF
--- a/apps/common/src/python/mediawords/solr/request.py
+++ b/apps/common/src/python/mediawords/solr/request.py
@@ -177,7 +177,8 @@ def merge_responses(mc_32_bit_collection: dict,mc_64_bit_collection: dict):
 
     docs = []
 
-    docs.extend(mc_32_bit_response.get("docs", [])).extend(mc_64_bit_response.get("docs", []))
+    docs.extend(mc_32_bit_response.get("docs", []))
+    docs.extend(mc_64_bit_response.get("docs", []))
 
     new_response.update({
         "response": {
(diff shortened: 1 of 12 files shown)


apps/common/src/python/mediawords/solr/request.py
```diff
@@ -152,6 +157,54 @@ def __solr_error_message_from_response(response: Response) -> str:
     return error_message
 
 
+def merge_responses(dict1: dict,dict2: dict):
```
esirK: This looks good @thepsalmist I haven't tested the code below but I would suggest we try something like this; I think it reduces some repetition. cc @kilemensi

```
def merge_responses(mc_32_bit_collection: dict,mc_64_bit_collection: dict):
    new_response = {}

    new_response.update(mc_32_bit_collection.get("responseHeader", {})) # Will prevent a KeyError

    mc_32_bit_response = mc_32_bit_collection.get("response", {})
    mc_64_bit_response = mc_64_bit_collection.get("response", {})

    num_found = mc_32_bit_response.get("numFound", 0) + mc_64_bit_response.get("numFound", 0)
    start_index = mc_32_bit_response.get("start", 0) + mc_64_bit_response.get("start", 0)

    docs = []

    docs.extend(mc_32_bit_response.get("docs", [])).extend(mc_64_bit_response.get("docs", []))

    new_response.update({
        "response": {
            "numFound": num_found,
            "start": start_index,
            "docs": docs,
        }
    })

    # facets
    mc_32_bit_facets = mc_32_bit_response.get("facets", {})
    mc_64_bit_facets = mc_64_bit_response.get("facets", {})

    count = mc_32_bit_facets.get("count", 0) + mc_64_bit_facets.get("count", 0)

    categories = {}

    if 'categories' in mc_32_bit_facets:
        if mc_32_bit_facets.get('categories', {}).get('buckets', []):
            categories.update({"buckets": mc_32_bit_facets.get('categories', {}).get('buckets', []).extend(mc_64_bit_facets.get('categories', {}).get('buckets', []))})

    new_response.update({
        "facets": {
            "count": count,
            "categories": categories,
        }
    })

    return new_response
```

apps/common/src/python/mediawords/solr/request.py
```diff
@@ -219,21 +270,38 @@ def solr_request(path: str,
...
-        request.set_header(name='Content-Type', value=content_type)
-        request.set_header(name='Content-Length', value=str(len(content_encoded)))
-        request.set_content(content_encoded)
-
+        results = []
+        for collection in collections:
+            abs_uri = furl(f"{solr_url}/{collection}/{path}")
+            abs_uri = abs_uri.set(params)
+            abs_url = str(abs_uri)
+            request = Request(method='POST', url=abs_url)
```
esirK, line 305: We can drop the `abs_url = str(abs_uri)` unless the `abs_url` is used somewhere else.
```
 request = Request(method='POST', url=str(abs_url))
```

apps/common/src/python/mediawords/solr/request.py
```diff
@@ -219,21 +270,38 @@ def solr_request(path: str,
...
+            request.set_header(name='Content-Type', value=content_type)
```
esirK, line 306: From [this](https://github.com/mediacloud/backend/blob/aws-fix/apps/common/src/python/mediawords/util/web/user_agent/request/request.py#L53), I think we can combine all the headers while creating the `Request` something like

```
headers = {
    'Content-Type': content_type
    'Content-Length': str(len(content_encoded)),
}

```
```
request = Request(method='POST', url=abs_url, headers=headers)
```

apps/common/src/python/mediawords/solr/request.py
thepsalmist: LGTM @esirK

